PR #1, opened 2019-01-10
Converted furyctl to new blocks structure

gitirabassi:
I didn't change much of your concurrent downloading, just adapted. Do you like this implementation?


lzecca78:
[Auto-Join bash logic](https://trello.com/c/BxmiIhGe/22-auto-join-bash-logic)

PR #5, opened 2019-04-18
Scaffold command

kandros:
add `scaffold` command to create base folders and autogenerate kustomize file form vendors katalog kustomize files

### DIFF
--- a/cmd/scaffold.go
+++ b/cmd/scaffold.go
@@ -41,15 +41,14 @@ var scaffoldCommand = &cobra.Command{
 }
 
 func scaffoldK8sManifests() {
-	dir := util.MustGetCurrentDir()
-	p := filepath.Join(dir, "manifests")
-	util.MustMkdir(p)
+	util.MustMkdirInCurrentDirectory("manifests")
 
 	type kustomizationFile struct {
 		Bases []string `yaml:"bases"`
 	}
 
-	absoluteVendorPath := filepath.Join(dir, "vendor")
+	currentDir := util.MustGetCurrentDir()
+	absoluteVendorPath := filepath.Join(currentDir, "vendor")
 	if _, err := os.Stat(absoluteVendorPath); os.IsNotExist(err) {
 		fmt.Println("no v endor folder found, run 'furyctl install'")
 		os.Exit(0)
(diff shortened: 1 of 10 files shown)


util/util.go
```diff
@@ -0,0 +1,59 @@
...
+func MustMkdirInCurrentDirectory(dirname string) {
+	dir := MustGetCurrentDir()
+	p := filepath.Join(dir, dirname)
+	MustMkdir(p)
+}
+
+func MustMkdir(dirname string) {
+	err := os.Mkdir(dirname, 0700)
+	if err != nil {
+		panic(err)
```
phisco: Would be nice to have idempotency, panicing prevents further steps to be performed.
At least don't panic if `os.IsExist(err)`.

kandros:
Gracefully handling and show instructions to recover instead

util/util.go
```diff
@@ -0,0 +1,61 @@
+package util
```
omissis: please try to avoid utility packages: https://dave.cheney.net/practical-go/presentations/qcon-china.html#_avoid_package_names_like_base_common_or_util

util/util.go
```diff
@@ -0,0 +1,61 @@
...
+*.tfstate
+*.backup
+			`)
+		}
+		os.Exit(0)
+	}
+	return nil
+}
+
+func FindBasesFromVendor(vendorPath string) (paths []string, err error) {
```
omissis: I would move this function to a more package with a more "speaking" name, such as `furyfile`

cmd/scaffold.go
```diff
@@ -0,0 +1,94 @@
...
+
+	content, err := yaml.Marshal(&kustomizationFile{Bases: relativeBases})
+	if err != nil {
+		panic(err)
+	}
+
+	util.SafeWriteFileOrExit(filepath.Join(currentDir, "manifests/kustomization.yml"), content)
+}
+
+func scaffoldAnsibleRules() {
```
omissis: this one and the functions below could be moved to a more specific `furyfile` package

cmd/scaffold.go
```diff
@@ -0,0 +1,94 @@
...
+func init() {
+	rootCmd.AddCommand(scaffoldCommand)
+}
+
+var scaffoldCommand = &cobra.Command{
+	Use:   "scaffold",
+	Short: "create required folders and files",
+	Long:  "scaffold a fury project creating required folders and files",
+	Run: func(cmd *cobra.Command, args []string) {
+		scaffoldTerraformModules()
```
omissis: I would also move these functions out of the `cmd` package(into `furyfile` maybe?) as they are not about command line infrastructure, they are about domain logic.

util/util_test.go
```diff
@@ -0,0 +1,20 @@
+package util
```
omissis: please use a separate package for testing, ie: `util_test`. this will allow you to better understand the exposed surface of your package

util/util_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestFindBasesFromVendor(t *testing.T) {
+	vendorPath := "../fixtures/vendor/"
```
omissis: I suggest using this approach to test fixtures: https://dave.cheney.net/2016/05/10/test-fixtures-in-go

angelbarrera92:
Uhmmm this PR seems to be too old. I wanted to do something with it. But i prefer to use something already built instead of make our own template engine.

Also i see this PR was focused only on "infra" repository structure. What about modules or distribution?

angelbarrera92:
[Repository template: Infrastructure](https://trello.com/c/Y6Vifilk/462-repository-template-infrastructure)

kandros:
clearly too old, closing

PR #9, opened 2019-10-28
Furyfile overridable "global versions" for roles/modules/bases

phisco:
I'm using the keys in the "versions" map as a prefix to decide if a module has to use the global version specified. I was thinking about using regexes, but sounded overkill.
You can find an updated and commented example in the Readme and a Furyfile in the root of the repo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,24 +14,26 @@ You can omit a type if you don't need any of its packages. An example Furyfile w
 would be like following:
 
 ```
+# all sections are optional
+
+# map of prefixes and versions used to force a specific version for all the matching roles/modules/bases
+versions:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -14,24 +14,26 @@ You can omit a type if you don't need any of its packages. An example Furyfile w
...
+  - name: aws/aws-kubernetes
 
 bases:
-  - name: monitoring/prometheus-operator
-    version: master
-  - name: monitoring/prometheus-operated
-    version: master
+  - name: monitoring
   - name: logging
+  # versions can be overridden if needed by specifying them for each package
```
angelbarrera92: Cool!

angelbarrera92:
LGTM!

PR #11, opened 2020-01-02
git via https

### LINKED ISSUE #10: Option to git clone via http(s)
Right now `furyctl install` downloads packages via git clone that uses ssh. In some environments the ssh use is banned. It would be nice if `furyctl` had a flag that would enable the clone/download via http(s) instead

angelbarrera92:
FIX #10 

I don't know if it is the right implementation, but have tested with a simple furyfile.yml file and it works.

@phisco, let me know if it works for you.

Once compiled, execute `furyctl install --htps`

Thanks!

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -22,7 +22,8 @@ import (
 
 const (
 	configFile              = "Furyfile"
-	repoPrefix              = "git@github.com:sighupio/fury-kubernetes"
+	httpsRepoPrefix         = "git::https://github.com/sighupio/fury-kubernetes"
+	sshRepoPrefix           = "git@github.com:sighupio/fury-kubernetes"
 	defaultVendorFolderName = "vendor"
 )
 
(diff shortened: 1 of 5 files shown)


cmd/config.go
```diff
@@ -22,7 +22,8 @@ import (
 
 const (
 	configFile              = "Furyfile"
-	repoPrefix              = "git@github.com:sighupio/fury-kubernetes"
+	httpsRepoPrefix         = "git::https://github.com/sighupio/fury-kubernetes"
```
ralgozino: furyctl prints the url when downloading a package, and using https you see this strange url that starts with `git::`, is a little bit misleading. Like this:

```shell
2020/01/03 09:39:53 downloading: git::https://github.com/sighupio/fury-kubernetes-on-premise.git//roles/sighup-base?ref=master -> vendor/roles/on-premise/sighup-base
```

What do you think of droping the `git::` when we print the url?

angelbarrera92: Also double `//` are showed. 
We can delete both if we want ;)

phisco:
LGTM, the logging could surely benefit some love.

angelbarrera92:
I think we can merge this change, then open an issue to improve logs. What's your opinion?

phisco:
Go for it @angelbarrera92

angelbarrera92:
> Go for it @angelbarrera92

Opened: https://github.com/sighupio/furyctl/issues/12

PR #13, opened 2020-01-25
added capability to handle fetch from public terraform registry

lzecca78:
Hi everyone 👋 !
as asked by @nutellinoit , i implemented in furyctl the feature to customize the yaml in order to support also public terraform modules from terraform registry. It works, please review the code and ask for everything is you see some lacks or mistakes!

### DIFF
--- a/Furyfile.yml
+++ b/Furyfile.yml
@@ -39,15 +39,18 @@ bases:
   - name: logging
     version: master
 
-tfRepos:
-  aws:
-    - url: https://github.com/terraform-aws-modules
-      label: ufficial-modules
-    - url: https://github.com/terraform-community-modules
-      label: community-modules
-  gcp:
-    - url: https://github.com/terraform-google-modules
-      label: ufficial-modules
-  azure:
-    - url: https://github.com/Azure
-      label: ufficial-modules
+provider:
+  roles: {}
+  bases: {}
+  modules:
+    aws:
+      - url: https://github.com/terraform-aws-modules
+        label: ufficial-modules
+      - url: https://github.com/terraform-community-modules
+        label: community-modules
+    gcp:
+      - url: https://github.com/terraform-google-modules
+        label: ufficial-modules
+    azure:
+      - url: https://github.com/Azure
+        label: ufficial-modules
--- a/Furyfile.yml
+++ b/Furyfile.yml
@@ -15,6 +15,12 @@ modules:
       name: aws
       label: ufficial-modules
     registry: true
+  - name: tf_aws_bastion_s3_keys
+    version: v2.0.0
+    provider:
+      name: aws
+      label: community-modules
+    registry: true
   - name: terraform-google-kubernetes-engine
     version: v6.2.0
     provider:
@@ -37,6 +43,8 @@ tfRepos:
   aws:
     - url: https://github.com/terraform-aws-modules
       label: ufficial-modules
+    - url: https://github.com/terraform-community-modules
+      label: community-modules
   gcp:
     - url: https://github.com/terraform-google-modules
       label: ufficial-modules
--- a/Furyfile.yml
+++ b/Furyfile.yml
@@ -11,18 +11,35 @@ modules:
   - name: aws/aws-kubernetes
   - name: terraform-aws-acm
     version: v2.5.0
-    provider: aws
+    provider:
+      name: aws
+      label: ufficial-modules
     registry: true
   - name: terraform-google-kubernetes-engine
     version: v6.2.0
-    provider: gcp
+    provider:
+      name: gcp
+      label: ufficial-modules
     registry: true
   - name: terraform-azurerm-aks
     version: v2.0
-    provider: azure
+    provider:
+      name: azure
+      label: ufficial-modules
     registry: true
 
 bases:
   - name: monitoring
   - name: logging
     version: master
+
+tfRepos:
+  aws:
+    - url: https://github.com/terraform-aws-modules
+      label: ufficial-modules
+  gcp:
+    - url: https://github.com/terraform-google-modules
+      label: ufficial-modules
+  azure:
+    - url: https://github.com/Azure
+      label: ufficial-modules
--- a/Furyfile.yml
+++ b/Furyfile.yml
@@ -11,6 +11,15 @@ modules:
   - name: aws/aws-kubernetes
   - name: terraform-aws-acm
     version: v2.5.0
+    provider: aws
+    registry: true
+  - name: terraform-google-kubernetes-engine
+    version: v6.2.0
+    provider: gcp
+    registry: true
+  - name: terraform-azurerm-aks
+    version: v2.0
+    provider: azure
     registry: true
 
 bases:
--- a/Furyfile.yml
+++ b/Furyfile.yml
@@ -13,7 +13,7 @@ modules:
     version: v2.5.0
     provider:
       name: aws
-      label: ufficial-modules
+      label: official-modules
     registry: true
   - name: tf_aws_bastion_s3_keys
     version: v2.0.0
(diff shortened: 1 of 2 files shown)


cmd/config.go
```diff
@@ -24,6 +24,7 @@ const (
 	configFile              = "Furyfile"
 	httpsRepoPrefix         = "git::https://github.com/sighupio/fury-kubernetes"
 	sshRepoPrefix           = "git@github.com:sighupio/fury-kubernetes"
+	registryRepoPrefix      = "git::https://github.com/terraform-aws-modules"
```
nutellinoit: Hi @lzecca78 , why only terraform-aws-modules? 

Should we add another parameter on furyfile like cloud with values like: "aws" "azure" "gcp" to search on respective GitHub organizations? And perhaps a hardcoded struct that contains which Github organizations are "official" for us.

Nice add btw :blush:

lzecca78: pretty simple @nutellinoit : this pr comes from the issue you wrote here :https://github.com/sighupio/fury-kubernetes-aws/issues/22 . That was related to the only terraform-aws-modules. Once it will be accepted, implement it into something that can be  customized for each cloud provider, is really simple btw. About the idea of many github organizations, imho at this time is overengineering. The xp approach that provides a really small step by step iteration flow, i believe is the right now in this situations. Once is ok, i will extend it pretty easily!

Furyfile.yml
```diff
@@ -11,6 +11,15 @@ modules:
...
     version: v2.5.0
+    provider: aws
+    registry: true
+  - name: terraform-google-kubernetes-engine
+    version: v6.2.0
+    provider: gcp
+    registry: true
+  - name: terraform-azurerm-aks
+    version: v2.0
+    provider: azure
```
lzecca78: @nutellinoit  now you can rest and dream without worries 🛌

cmd/config.go
lzecca78: i close this because it refers to outdated code :)

Furyfile.yml
lzecca78: outdated code

Furyfile.yml
```diff
@@ -10,8 +9,45 @@ roles:
 modules:
   - name: aws/aws-vpc
   - name: aws/aws-kubernetes
+  - name: terraform-aws-acm
+    version: v2.5.0
+    provider:
+      name: aws
+      label: ufficial-modules
```
angelbarrera92: Please, replace all `ufficial` to `official`

Furyfile.yml
```diff
@@ -10,8 +9,45 @@ roles:
...
+      name: azure
+      label: ufficial-modules
+    registry: true
 
 bases:
   - name: monitoring
   - name: logging
     version: master
+
+tfRepos:
```
angelbarrera92: What about changing `tfRepos` to `provider.module`?

I mean, what if tomorrow we need to do the same with roles or katalog (kustomize)?

The change will looks like:

Before:
```yaml
tfRepos:
  aws:
    - url: https://github.com/terraform-aws-modules
      label: ufficial-modules
```

To

```yaml
provider:
  modules:
    aws:
      - url: https://github.com/terraform-aws-modules
        label: ufficial-modules
  roles: {} #TBD
  katalog: {} #TBD
```

lzecca78: could be an idea, but is better to implement it as asoon as we will have the real need (the same that push me to deepen this development). The risk of the overengineer is to build something that won't match the need, once the need will appear. Just to explain it better: try to consolidate upfront can expose a wrong abstraction of the other 2 (roles and katalog) about what now we have no information on how could be a future implementation.

Furyfile.yml
lzecca78: italianglish prevails over english, my typo

Furyfile.yml
lzecca78: i applied the change you've asked, give a look 👀

angelbarrera92: Now looks very good!!! Thanks @lzecca78 !

lzecca78:
[Furyctl feature for handling injection of external registries/url](https://trello.com/c/DkwswM1C/55-furyctl-feature-for-handling-injection-of-external-registries-url)

cmd/config.go
```diff
@@ -29,26 +29,57 @@ const (
...
+// modules:
+//   aws
+//    - uri: https://github.com/terraform-aws-modules
+//      label: ufficial-modules
+type ProviderKind map[string][]RegistrySpec
+
+//RegistrySpec contains the couple uri/label to identify each tf new repo declared
+type RegistrySpec struct {
+	BaseURI string `mapstructure:"url"`
+	Label   string `yaml:"label"`
```
phisco: mixing yaml and mapstructure here? is it working fine?

cmd/config.go
```diff
@@ -92,13 +129,113 @@ func (f *Furyconf) Parse() ([]Package, error) {
...
+func newDir(folder, kind, name string, registry bool, provider ProviderSpec) *DirSpec {
+	return &DirSpec{
+		VendorFolder: folder,
+		Kind:         kind,
+		Name:         name,
+		Registry:     registry,
+		Provider:     provider,
+	}
+}
+
```
phisco: no need for the constructor if the DirSpec structure is public and the constructor is not enforcing anything.

cmd/config.go
```diff
@@ -92,13 +129,113 @@ func (f *Furyconf) Parse() ([]Package, error) {
...
+	return &URLSpec{
+		Registry:       registry,
+		Prefix:         prefix,
+		Blocks:         blocks,
+		DotGitParticle: dotGitParticle,
+		Kind:           kind,
+		Version:        version,
+	}
+}
+
```
phisco: Same goes here, no need for the constructor

cmd/config.go
```diff
@@ -92,13 +129,113 @@ func (f *Furyconf) Parse() ([]Package, error) {
...
+func (n *URLSpec) strategy() string {
```
phisco: maybe could benefit of a better name

cmd/config.go
```diff
@@ -92,13 +129,113 @@ func (f *Furyconf) Parse() ([]Package, error) {
...
+	var url string
+	if n.Registry {
+		url = fmt.Sprintf("%s/%s%s?ref=%s", n.Prefix, n.Blocks[0], n.DotGitParticle, n.Version)
+	} else {
+		url = n.internalBehaviourURL()
+	}
+	return url
+}
+
+func (n *URLSpec) internalBehaviourURL() string {
```
phisco: maybe could benefit of a better name

cmd/config.go
lzecca78: you're right, i only create the constructor for a possible enhancement that could be done in the future, in this way the environment is already setup!

cmd/config.go
lzecca78: you're right, i only create the constructor for a possible enhancement that could be done in the future, in this way the environment is already setup!

cmd/config.go
lzecca78: i know  that is not the best name, please feel free to change with one better!

cmd/config.go
lzecca78: i know that is not the best name, please feel free to change with one better!

cmd/config.go
lzecca78: is working, probably should be better to put both as mapstructure? what do you think?

jnardiello:
See: https://github.com/sighupio/fury-kubernetes-aws/issues/22

angelbarrera92:
> See: [sighupio/fury-kubernetes-aws#22](https://github.com/sighupio/fury-kubernetes-aws/issues/22)

Due to: https://github.com/sighupio/fury-kubernetes-aws/issues/22#issuecomment-581129931
We can continue with this PR. What is the state?

lzecca78:
I guess is waiting to be merged

angelbarrera92:
> I guess is waiting to be merged

Ok, i take care of it ;) Thanks @lzecca78 !

PR #16, opened 2020-02-15
fix depth subpath bug

lzecca78:
Due to a bug discovered during the provisioning of module velero with @simonemex  and @lnovara   that has. a subdir of multiple path, here it is the patch for this bug. Now the function will handle whatever hierarchy level of subdir will exist in the repo

### DIFF
--- a/Furyfile.yml
+++ b/Furyfile.yml
@@ -5,12 +5,6 @@ versions:
 roles:
     - name: aws/etcd
     - name: aws/kube-control-plane
-    - name: dr/velero/velero-base
-      version: feature/deprecate-and-update
-    - name: dr/velero/velero-aws
-      version: feature/deprecate-and-update
-    - name: dr/velero/velero-restic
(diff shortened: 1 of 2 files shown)


lzecca78:
hi @ralgozino , you are right, is my vim configuration that is set in that way. Is pretty the usual configuration, but if is something that have to adapted to a standard, just let me know and i will change it  👍

angelbarrera92:
> Thanks @lzecca78 ! I'm seeing a lot of format changes in the diff (changed from 2 to 4 spaces identation), I think w

I was working on it but there is not ready yet. So feel free to continue the PR

lzecca78:
Anyway I made a search and the standard in yaml Is 2 spaces so I changed my vim setup to match this standard.

angelbarrera92:
LGTM! Waiting for one more feedback, @ralgozino  or @phisco

PR #20, opened 2020-12-07
now the download can be repeated without errors

lzecca78:
Hi everyone 👋 !
As @nutellinoit  said in a previous issue, there is currently a problem with public modules that embed also the `.git` subfolders. 
The current PR resolves the following misbehaviors: 

1. errors when you want to re-run a furyctl vendor multiple times (more than 1 actually :D ) 
2. remove automatically `.git` subfolder from the public modules that contain the `.git` subfolder

Basically now the final behavior is the same of the fury module. 
Let me know if everything looks good for you!

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"fmt"
 	"github.com/sirupsen/logrus"
 	"os"
 	"runtime"
@@ -67,7 +68,6 @@ func get(src, dest string, mode getter.ClientMode) error {
 
 	logrus.Debugf("complete url downloading: %s -> %s", src, dest)
 
-	humanReadableDownloadLog(src, dest)
 
 	pwd, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


lnovara (approved):

nutellinoit (approved):
LGTM 🎉

cmd/download.go
```diff
@@ -80,8 +80,32 @@ func get(src, dest string, mode getter.ClientMode) error {
...
+		removeContents(gitFolder)
+	}
+
+
+	/*	gitFolder := fmt.Sprintf("%s/.git",dest)
+
+		if _, err := os.Stat(gitFolder); !os.IsNotExist(err) {
+			logrus.Infof("%s found, removing it!",gitFolder)
+			removeContents(gitFolder)
+		}*/
```
angelbarrera92, line 106: Why is this code block commented?

cmd/download.go
```diff
@@ -80,8 +80,32 @@ func get(src, dest string, mode getter.ClientMode) error {
...
+		humanReadableDownloadLog(src, dest)
+		err = client.Get()
+		logrus.Infof("removing %s",gitFolder)
+		removeContents(gitFolder)
+	}else {
+		humanReadableDownloadLog(src, dest)
+		err = client.Get()
+		logrus.Infof("removing %s",gitFolder)
+		removeContents(gitFolder)
+	}
```
angelbarrera92, line 98: I think we can avoid dup code using:

```go
if _, err := os.Stat(dest); !os.IsNotExist(err) {
		logrus.Infof("%s already exists! removing it",dest)
		removeContents(dest)
}
err = client.Get()
logrus.Infof("removing %s",gitFolder)
removeContents(gitFolder)
```

Am i right? What do you think about this approach?

cmd/download.go
```diff
@@ -102,3 +126,12 @@ func humanReadableDownloadLog(src string, dest string) {
...
 }
+
+
+func removeContents(dir string) error {
+	err:= os.RemoveAll(dir)
+		if err != nil {
+			return err
+		}
+	return nil
+}
```
angelbarrera92, line 137: Missing carriadge return.

angelbarrera92 (changes requested):
Minor coding style comments, anyway, it should work. Take a look at them @lzecca78.

cmd/download.go
lzecca78: yes, this should work! Was a rusty PR i know! my intention was to do a refactoring today, but you anticipate me! I will test it with the comment removal and rename to a function! thanks @angelbarrera92 !

cmd/download.go
lzecca78: because i left it there :D,removing it!

cmd/download.go
lzecca78: i don't get this one @angelbarrera92  🤔

angelbarrera92: A new line at the end of the file :)

angelbarrera92 (approved):
LGTM!

angelbarrera92 (approved):
LGTM!

angelbarrera92:
The new release contains this bugfix. Thanks @lzecca78 !!!

PR #22, opened 2020-12-09
WIP: reverse tunnel design document

gabrielelana:
Still working on it but if you want to keep an eye and make some comments feel free to do so ;-)

### DIFF
--- a/docs/design/002-reverse-tunnel.md
+++ b/docs/design/002-reverse-tunnel.md
@@ -1,0 +1,193 @@
+# LGTM (Looks Good To Me)
+
+- [ ] Ángel Barrera Sánchez
+- [ ] Giovanni Laieta
+- [ ] Luca Novara
+- [ ] ???
+
+# Status+
+
+_Draft_: the paragraphs that are prefixed or suffixed with `...`
+should be considered scratchpads for ideas that need further thinking
+and structure.
+
+# Problem
+
+Every Fury customer have different security policies (some are
+stricter than others) and different implementation of those
+policies.
+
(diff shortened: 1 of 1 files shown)


angelbarrera92:
Far from sound rude, I would like to propose using wireguard instead of reinventing ssh tunnelings (that are unprotected...). Wireguard cipher the connectivity and it enables the same as described in the document.

I mean... all the agents could be wireguard clients. The bastion is just wireguard servers... We are going to end up implementing something like https://github.com/subspacecloud/subspace

The idea is clever, and we really need it. What i don't want is to reimplement the wheel

gabrielelana:
@angelbarrera92 since it's a @glaieta proposal I'll let him check on your proposal. @glaieta let us know what do you think.

glaieta:
> Far from sound rude, I would like to propose using wireguard instead of reinventing ssh tunnelings (that are unprotected...). Wireguard cipher the connectivity and it enables the same as described in the document.
> 
> I mean... all the agents could be wireguard clients. The bastion is just wireguard servers... We are going to end up implementing something like https://github.com/subspacecloud/subspace
> 
> The idea is clever, and we really need it. What i don't want is to reimplement the wheel

Too long to explain....I start to think to use a vpn with wireguard.... but wireguard is a VPN and some customers see a VPN like a method to connect inside their network. SSH open connection is see as connection to an host and not an intrusion. For example Tenaris accept ssh reverse tunnel but doesn't accept a VPN.
If you want I can explain you by call conference.

gabrielelana:
Outdated, superseded by `fury-connect` CVI

PR #23, opened 2021-01-18
Little flag to filter updates in Furyfile

nutellinoit:
Hi team!

I added a little flag on the furyctl vendor command ~`--filter`~ `--prefix` to add a prefix on the modules i want to update, eg: ~`furyctl vendor --filter dr/velero`~  `furyctl vendor --prefix dr/velero`.

This can useful when we do stack upgrades.

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -89,20 +89,38 @@ func (f *Furyconf) Validate() error {
 }
 
 // Parse reads the furyconf structs and created a list of packaged to be downloaded
-func (f *Furyconf) Parse() ([]Package, error) {
+func (f *Furyconf) Parse(filter string) ([]Package, error) {
 	pkgs := make([]Package, 0, 0)
 	// First we aggreggate all packages in one single list
 	for _, v := range f.Roles {
 		v.kind = "roles"
-		pkgs = append(pkgs, v)
+		if len(filter)>0 {
+			if strings.HasPrefix(v.Name, filter) {
+				pkgs = append(pkgs, v)
+			}
+		}else{
+			pkgs = append(pkgs, v)
+		}
 	}
 	for _, v := range f.Modules {
 		v.kind = "modules"
-		pkgs = append(pkgs, v)
+		if len(filter)>0 {
+			if strings.HasPrefix(v.Name, filter) {
+				pkgs = append(pkgs, v)
+			}
+		}else{
+			pkgs = append(pkgs, v)
+		}
 	}
 	for _, v := range f.Bases {
 		v.kind = "katalog"
-		pkgs = append(pkgs, v)
+		if len(filter)>0 {
+			if strings.HasPrefix(v.Name, filter) {
+				pkgs = append(pkgs, v)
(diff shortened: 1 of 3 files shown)


phisco:
Is hasprefix enough? What about regexes? Mainly to be able to do an "or" if needed as I don't think "something/.*/whatever" would be a realistic use case in favor of regexes.

Otherwise LGTM

angelbarrera92:
> Is hasprefix enough? What about regexes? Mainly to be able to do an "or" if needed as I don't think "something/.*/whatever" would be a realistic use case in favor of regexes.
> 
> Otherwise LGTM

Agree with Phillipe!

nutellinoit:
I would like to start "easy", in my day to day for this type of activity i don't need regex capabilities, i need only to reduce furyctl vendor scope when updating vendor folder.

angelbarrera92:
> I would like to start "easy", in my day to day for this type of activity i don't need regex capabilities, i need only to reduce furyctl vendor scope when updating vendor folder.

In this case, I suggest renaming the flag. Instead of `f` | `filter`, something like `prefix`. Then we can implement the regex feature later using the `f` | `filter` flag. 

What do you think?

nutellinoit:
> > I would like to start "easy", in my day to day for this type of activity i don't need regex capabilities, i need only to reduce furyctl vendor scope when updating vendor folder.
> 
> In this case, I suggest renaming the flag. Instead of `f` | `filter`, something like `prefix`. Then we can implement the regex feature later using the `f` | `filter` flag.
> 
> What do you think?

Sure @angelbarrera92 , i will do it now

nutellinoit:
Done, instead of filter i added `--prefix` with the `-P` short notation

angelbarrera92:
> Done, instead of filter i added `--prefix` with the `-P` short notation

Can you take a look at the linter? https://ci.sighup.io/sighupio/furyctl/159/1/3

nutellinoit:
> > Done, instead of filter i added `--prefix` with the `-P` short notation
> 
> Can you take a look at the linter? https://ci.sighup.io/sighupio/furyctl/159/1/3

Done!

angelbarrera92 (approved):
LGTM!

cmd/config.go
```diff
@@ -89,20 +89,38 @@ func (f *Furyconf) Validate() error {
...
+		}
 	}
 	for _, v := range f.Bases {
 		v.kind = "katalog"
-		pkgs = append(pkgs, v)
+		if len(prefix) > 0 {
+			if strings.HasPrefix(v.Name, prefix) {
+				pkgs = append(pkgs, v)
+			}
+		} else {
```
Rampage1xx, line 121: same issue

Rampage1xx (commented):
my feedbacks

Rampage1xx:
LGTM

ralgozino (approved):
lgtm

angelbarrera92:
Released at v0.2.5 :)

PR #24, opened 2021-01-20
CVI 001 - Cluster Self Provisioning

angelbarrera92:
This PR contains the MVP of the self-serving / self-provisioning cluster.

TODO: Complete the PR description
meanwhile, add your review/comments as PR comments

---
Missing. Spotted from the review. Thank you all :)

- [X] `destroy` subcommands requires confirmation.
- [x] consider renaming the `update` command to `apply`.
- [x] Implement `init --reset` flag to recreate the terraform project.
- [x] Add alternative way to pass the `--token`. Add `FURYCTL_TOKEN`.
- [x] Improve terraform code to pin terraform provider versions.
- [x] Implement `furyctl {bootstrap,cluster} template` to output a sample yml to let the user modify it easily.
- [x] Move the backend configuration option from the `.conf` file to `backend.tf`.
- [x] Fix imperative variables on `aws` bootstrap provisioner
- [x] Add a `secrets` directory along with `.gitattributes` to encrypt that folder out of the box if the client has `git-crypt` installed.
- [x] Clean (somehow) the env variables starting by TF_VAR as they can affect the current behaviour of the `furyctl`
- [X] in the bootstrap phase, add an option to tweak the number of bastions create (which could also be 0)

---
Out of scope for the first release/MVP
- [ ] Consider a command that creates a `cluster.yml` file from the `bootstrap` output.

---
Out of `furyctl` scope:
- [ ] Change the cluster provisioner to accept a file path to read the public key instead of pasting the public key content.
- [ ] in the bootstrap phase, add an option to skip creation of the VPC and use a customer-provided one
---

Discarded:

- [ ] preflight checks (maybe terraform validate¿?) before return 0 on `init` commands. (Check the executor parameters)

Reason: `furyctl {bootstrap,cluster} init` performs `terraform init` under the hood. It validates "enough" data to don't reinvent the wheel here. Example with an incorrect `backend configuration`:

<details>
  <summary>Click to expand!</summary>

```bash

furyctl bootstrap init --reset
WARN[0000] Cleaning up the workdir                      
WARN[0000] Removing /Users/angelbarrerasanchez-sighup/work/sighup/furyctl/demo/./bootstrap directory 
WARN[0000] could not find terraform executable          
INFO[0002] [INFO] running Terraform command: /Users/angelbarrerasanchez-sighup/work/sighup/furyctl/demo/bootstrap/bin/terraform init -no-color -force-copy -input=false -lock-timeout=0s -backend=true -get=true -get-plugins=true -lock=true -upgrade=false -verify-plugins=true 
⣷ Initializing terraform project ERRO[0017] error while running terraform init in the project dir: 2021/01/28 17:02:59 [DEBUG] Using modified User-Agent: Terraform/0.12.29 HashiCorp-terraform-exec/0.11.0

Error: Failed to get existing workspaces: S3 bucket does not exist.

The referenced S3 bucket must have been previously created. If the S3 bucket
was created within the last minute, please wait for a minute or two and try
again.

Error: NoSuchBucket: The specified bucket does not exist
        status code: 404, request id: B3ADF527EBD883FD, host id: Gr3FnYLOCJG8myuidA7hztjrFh0HyjM04yndoMKcAKRzWeHeX7+o4ujgZFpmZy78rpwhbBg047A=


 
Error: 2021/01/28 17:02:59 [DEBUG] Using modified User-Agent: Terraform/0.12.29 HashiCorp-terraform-exec/0.11.0

Error: Failed to get existing workspaces: S3 bucket does not exist.

The referenced S3 bucket must have been previously created. If the S3 bucket
was created within the last minute, please wait for a minute or two and try
again.

Error: NoSuchBucket: The specified bucket does not exist
        status code: 404, request id: B3ADF527EBD883FD, host id: Gr3FnYLOCJG8myuidA7hztjrFh0HyjM04yndoMKcAKRzWeHeX7+o4ujgZFpmZy78rpwhbBg047A=




Usage:
  furyctl bootstrap init [flags]

Flags:
  -c, --config string    Bootstrap configuration file path (default "bootstrap.yml")
  -h, --help             help for init
      --reset            Forces the re-initialization of the project. It deletes the content of the workdir recreating everything
  -t, --token string     GitHub token to access enterprise repositories. Contact sales@sighup.io
  -w, --workdir string   Working directory to create and place all project files. Must not exists. (default "./bootstrap")

Global Flags:
      --debug     Enables furyctl debug output
  -d, --disable   Disable analytics

ERRO[0017] 2021/01/28 17:02:59 [DEBUG] Using modified User-Agent: Terraform/0.12.29 HashiCorp-terraform-exec/0.11.0

Error: Failed to get existing workspaces: S3 bucket does not exist.

The referenced S3 bucket must have been previously created. If the S3 bucket
was created within the last minute, please wait for a minute or two and try
again.

Error: NoSuchBucket: The specified bucket does not exist
        status code: 404, request id: B3ADF527EBD883FD, host id: Gr3FnYLOCJG8myuidA7hztjrFh0HyjM04yndoMKcAKRzWeHeX7+o4ujgZFpmZy78rpwhbBg047A=
```

</details>


nutellinoit (commented):
Hi @angelbarrera92, here some comments after testing the provision process

### Common suggestions

1) I want to see what `Applying terraform project` is doing
2) (personal opinion) I don't like the "update" command, i prefer an "apply" command
3) I would like to trigger init command multiple times **before** the first terraform apply, to check inside the folder if everything is as i expect. Something like `furyctl bootstrap/cluster init --reset` can do the job.
4) Add on the documentation how to create a github token and which scopes are needed
5) Possibility to use an env var like `GITHUB_TOKEN` or something similar
6) Missing providers pinning best practice on `main.tf`, also missing default region setting

### Bootstrap process

1) It's not easy to copy/paste from the doc, to create a bootstrap.yml file. I would like a `furyctl bootstrap template` command with an example.
2) It's not clear what vpn.subnetCIDR should contain. As far as i understood, must be one of the previous created subnets
3) The s3 backend is not written inside the backend.tf file, but on some other .conf file. I want to be able to run the terraform project without furyctl
4) `/xxx/xxx/github.com/sighupio/furyctl/demo/bootstrap/bin/terraform apply -no-color -auto-approve -input=false -lock=true -parallelism=10 -refresh=true -var name=unbelprogetto -var network_cidr=10.0.0.0/16 -var public_subnetwork_cidrs=["10.0.1.0/24","10.0.2.0/24","10.0.3.0/24"] -var private_subnetwork_cidrs=["10.0.101.0/24","10.0.102.0/24","10.0.103.0/24","10.0.104.0/24"] -var vpn_subnetwork_cidr=10.0.104.0/24 -var vpn_ssh_users=["nutellinoit","Rampage1xx"]`
   
   I don't like the imperative approach here, i would like to have all these variables on a tfvar file. This way i can check them during the init phase. This unlocks also the possibility to run terraform inside the bootstrap folder.
5) Secrets should be in a secrets folder, eg furyagent.yml
6) I would like to have a default `.gitattributes` file with all the secrets folders configured for git-crypt (plug and play) , or a command to to that


### Cluster bootstrap process

1) To be able to reduce errors, i would like to use some `furyctl bootstrap output cluster eks` command that creates for me a cluster.yml file with
   the outputs from the bootstrap project.
2) dmzCIDRRange cannot be a single CIDR, we need an array. 
3) I don't like the approach of pasting the public key, i would like to use a filepath and read the content with terraform
4) nodePools version cannot be empty or null, otherwise you get an `Error: Your query returned no results. Please change your search criteria and try again.`

### Questions

1) Can i manage aws-auth? I don't want the installer to do that for me

Awesome job btw! :tada:

nutellinoit:
PS:

1) Destroy command is not asking for permission to proceed 😱 
2) Why we are using modules from https/github and not locally via `vendor` folder?

angelbarrera92:
First of all, thank you @nutellinoit for taking the time to test this PR :)
I'll try to answer some points:

> I want to see what Applying terraform project is doing
By default, you don't see terraform output. We wanted to hide what's under the roof. Anyway, you've got two different ways to see terraform logs:
1. As you know, furyctl creates a directory with all the terraform project. Inside it, there is a `logs` directory with all the logs. You can open a new terminal then tail that file.
2. If you run the same command with `--debug`, terraform output is available during the execution of the command.

The idea is to (try) to forget about terraform, at the same time, if we need to look or understand what is happening, there is a way to inspect it :)

Don't know if it makes sense.

angelbarrera92:
> (personal opinion) I don't like the "update" command, i prefer an "apply" command
Ye... I didn't want to be "so" terraform. I copied some of the `kops` experience. Anyway, if more people want it to be renamed, I can do it :)

angelbarrera92:
> I would like to trigger init command multiple times before the first terraform apply, to check inside the folder if everything is as i expect. Something like furyctl bootstrap/cluster init --reset can do the job.

🤔 it could be a good idea. What do you expect from this `flag`? A complete remove and recreate directory?
It could be a good idea to discard "possible" manual modification. Does it make sense to you?

angelbarrera92:
> Add on the documentation how to create a github token and which scopes are needed
About 🔝 
The token is something we are going to create for the paid customers. Anyway, it could make sense to document it somehow. Just take in mind that the token is something a `paid customer` will receive after payment.

angelbarrera92:
> Possibility to use an env var like GITHUB_TOKEN or something similar

angelbarrera92:
> Missing providers pinning best practice on main.tf, also missing default region setting

That's something we have to work on, for sure. Thanks for popping it up

angelbarrera92:
> It's not easy to copy/paste from the doc, to create a bootstrap.yml file. I would like a furyctl bootstrap template command with an example.

Yes! I was thinking about it too. I'll do, it makes totally sense.

angelbarrera92:
> It's not clear what vpn.subnetCIDR should contain. As far as i understood, must be one of the previous created subnets

This is something described and available in the aws bootstrap provisioner docs: http://kfd-docs-feature-update-furyctl.surge.sh/docs/cli-reference/furyctl/provisioners/aws-bootstrap/#configuration

It is the subnet CIDR for the VPN. It must be different from the two networks the VPN is connecting

lzecca78:
>(personal opinion) I don't like the "update" command, i prefer an "apply" command 

I am agree with @nutellinoit , `update` is something that doesn't match with the action, so should be better `apply` because it is the real thing that is gonna happen after have executed that command :D

angelbarrera92:
> The s3 backend is not written inside the backend.tf file, but on some other .conf file. I want to be able to run the terraform project without furyctl

Ok, you can still run the terraform code with the created `.conf` file. Anyway, I agree moving that part to the backend.tf file.

angelbarrera92:
> I don't like the imperative approach here, i would like to have all these variables on a tfvar file. This way i can check them during the init phase. This unlocks also the possibility to run terraform inside the bootstrap folder.

Ye... I did imperative in the bootstrap process, but we can do as furyctl cluster does, creating a tfvar file. It makes sense to have it aligned.

angelbarrera92:
> Secrets should be in a secrets folder, eg furyagent.yml

> I would like to have a default .gitattributes file with all the secrets folders configured for git-crypt (plug and play) , or a command to to that

Yes! I'll do.

lzecca78:
>Possibility to use an env var like GITHUB_TOKEN or something similar

I would rather prefer something like GITHUB_TOKEN or CVS_TOKEN  (control version system) then FURYCTL_TOKEN, because in this case is the token that you need to interact with github not with something related to furyctl.

angelbarrera92:
> To be able to reduce errors, i would like to use some furyctl bootstrap output cluster eks command that creates for me a cluster.yml file with the outputs from the bootstrap project.

🤔 It could be a good idea not sure how to tackle it, but for sure makes sense

angelbarrera92:
> dmzCIDRRange cannot be a single CIDR, we need an array.

In that case, we need to start a new discussion about this parameter. Currently, the cluster uses the fury-eks-installer witch uses a standard interface. 

https://github.com/sighupio/fury-eks-installer/blob/master/modules/eks/variables.tf#L21-L24

angelbarrera92:
> I don't like the approach of pasting the public key, i would like to use a filepath and read the content with terraform

Makes sense!

lzecca78:
>I don't like the approach of pasting the public key, i would like to use a filepath and read the content with terraform

Totally agree. Moreover, can be used a reference to github users array, as was done for the bootstrap.yaml.

angelbarrera92:
> nodePools version cannot be empty or null, otherwise you get an Error: Your query returned no results. Please change your search criteria and try again.

🤔 it should be addressed in the installer first: https://github.com/sighupio/fury-eks-installer

angelbarrera92:
> Can i manage aws-auth? I don't want the installer to do that for me

You'll. We have to release a new version of the eks installer, then enable it in `furyctl`. So yes, you will

angelbarrera92:
> Destroy command is not asking for permission to proceed 😱

TOTALLY AGREE! TODO with some urgency :)

lzecca78:
> To be able to reduce errors, i would like to use some furyctl bootstrap output cluster eks command that creates for me a cluster.yml file with the outputs from the bootstrap project.

As i suggested the best way should be tagging the resources created by the bootstrap and then, reference them automatically in the cluster.yaml without having to fill it in granular way.

angelbarrera92:
> Why we are using modules from https/github and not locally via vendor folder?

I thought it was easier that way. Buy maybe we can consider the other way (vendor).

angelbarrera92:
> > Possibility to use an env var like GITHUB_TOKEN or something similar
> 
> I would rather prefer something like GITHUB_TOKEN or CVS_TOKEN (control version system) then FURYCTL_TOKEN, because in this case is the token that you need to interact with github not with something related to furyctl.

Today we are using GITHUB that's true. A client doesn't have to know where the module comes from.
Also, we can host the modules under a private fileserver with other auth methods. This is why I think it would be better if we name the env variable as `FURYCTL_TOKEN`.

angelbarrera92:
> > I don't like the approach of pasting the public key, i would like to use a filepath and read the content with terraform
> 
> Totally agree. Moreover, can be used a reference to github users array, as was done for the bootstrap.yaml.

This will require to modify the installer. We can talk about it along with the multiple cidr ranges feature

angelbarrera92:
From @lzecca78 , I've got:

> 1. During the bootstrap init step the procedure throws an error if you already have TF_VAR* environment variables defined globally

Not sure how to tackle this problem, but it is a problem for sure. Adding to the list

> 2. During the bootstrap yaml setup is not evidenced which lines need to be changed manually and are not explained with what needs to be changed.

Ye, maybe we have to improve the docs and it would be better once we completed the `furyctl cluster template stuff`.

> During the bootstrap update step the procedure creates the `bootstrap` directory even if the remote state bucket doesn’t exist.

Maybe we can add some kind of pre-flight checks. 🤔 how to do it?

lzecca78:
>Maybe we can add some kind of pre-flight checks. 🤔 how to do it?

Is not complicated imho. Depending on the adapter (s3 or others) we can implement providers to check if that bucket exists or not in the preflight check, and imho, the same provider can takes in charge the task of creating it if is missing and if the user want (a boolean value that allow the automation of this task)

lzecca78:
> Ye, maybe we have to improve the docs and it would be better once we completed the furyctl cluster template stuff.

about this i suggest to have a `seed.yml` file in which the user put all the stuff needed but the `template` flag creation, so all the information will be there in a single point

nikever:
Hello, 

I add my 2 cents. I will focus more on the overall design rather than implementation details and small improvements. The suggestions made by @nutellinoit and @lzecca78 were already super helpful in this last part.

**What I like**

- **Automation**. It automates some cumbersome parts of the delivery part.
- **Standardization**. It enforces a standard way of the delivery for the initial common parts. 
- **Simplicity**. Now it has some rough edges that can be improved, but overall the process seems very smooth. 

**What I don't like**

- **Terraform code is hidden**.  The terraform code has a secondary role in documenting the infrastructure; the new documentation becomes the `boostrap.yml` and `cluster.yml`. However, the logic of the module "hides" the output terraform code of the infrastructure.
- **Increase abstraction**. We add another layer of abstraction; now we have terraform -> terraform modules -> fury modules -> furyctl logic. This layer is yet another thing to maintain and a new possible source of bugs.
- **Not DRY**. It doesn't solve (yet) one of the problems that terraform has: code repetition across multiple environments. I have to do `furyctl bootstrap` `furyctl cluster` across various environments to get 98% of the same code across all of them.

nikever:
And P.S. Great work Angel as always 👏

angelbarrera92:
@nutellinoit About:

>  Change the cluster provisioner to accept a file path to read the public key instead of pasting the public key content.
Im having mixed feelings with this action. I don't know if we must implement this "additional" logic on top of the installer.

I mean...

The `eks` installer: https://github.com/sighupio/fury-eks-installer/tree/master/modules/eks#inputs has the ssh public key as an RSA key instead of a file path.

Then furyctl uses the installer under the hood exposing the variables as a yaml file (1:1).

*data/provisioners/cluster/eks/main.tf*
```hcl
module "fury" {
  source = "github.com/sighupio/fury-eks-installer//modules/eks?ref=v1.2.0"

  cluster_name    = var.cluster_name
  cluster_version = var.cluster_version
  network         = var.network
  subnetworks     = var.subnetworks
  dmz_cidr_range  = var.dmz_cidr_range
  ssh_public_key  = var.ssh_public_key
  node_pools      = var.node_pools
  tags            = var.tags
}
```

IMHO, adding this (simple) logic on `furyctl` covers a problem on the installer. So maybe we have to reconsider deprecate that field/attribute on the installer appending a new one: `ssh_public_key_path` then incorporate that change in `furyctl`. 

Let me know if you didn't understand it.

lnovara (commented):
Besides all the things already highlighted by @nutellinoit and @lzecca78, I would like to add the following:
  - avoid using a shared `sighup` user for the bastion (even though with multiple SSH keys)
  - in the bootstrap phase, add an option to skip creation of the VPC and use a customer-provided one
   - in the bootstrap phase, add an option to tweak the number of bastions create (which could also be 0)
   
I've mixed feelings about dumping the terraform code and allowing to fiddle with it, I think the aim of this work is to create an opinionated way to create Kubernetes clusters. Everything else should be handled with a separate Terraform project with a separate state.

Also, what about adding an `apiVersion` to the YAML files? I agree that right now is just overkill but in the future might well be put to good use.

nutellinoit:
I was also thinking, can we put some version pinning on the underlying modules used by furyctl? For example, fury-eks-installer v1.2.0 somewhere.

angelbarrera92:
> I was also thinking, can we put some version pinning on the underlying modules used by furyctl? For example, fury-eks-installer v1.2.0 somewhere.

Yes, we are already doing it. Currently, im updating eks provisioner :)

angelbarrera92:
> you will

I've implemented this feature :)

You can figure out how to use it by running the following command:

```bash
$ furyctl cluster template --provisioner eks
kind: Cluster
metadata:
  name: 'nameOfTheProject # Used to identify your resources'
  labels:
    boolLabel: true
    yourLabel: 'yourValue # You Can add as much labels as you required. You can add
      bool, string, number values'
spec:
  version: '1.18 # EKS Control plane version'
  network: 'vpc-id1 # Identificator of the VPC'
  subnetworks:
  - 'subnet-id1 # Identificator of the subnets'
  dmzCIDRRange: 10.0.0.0/8. Required. Network CIDR range from where cluster control
    plane will be accessible
  sshPublicKey: sha-rsa 190jd0132w. Required. Cluster administrator public ssh key.
    Used to access cluster nodes.
  nodePools:
  - name: my-node-pool. Required. Name of the node pool
    version: 1.18. Required. null to use Control Plane version.
    minSize: 0
    maxSize: 1
    instanceType: t3.micro. Required. AWS EC2 isntance types
    maxPods: 110
    volumeSize: 50
    labels:
      environment: 'dev. # Node labels. Use it to tag nodes then use it on Kubernetes'
    taints:
    - key1=value1:NoSchedule. As an example
    subnetworks:
    - subnet-1
    - '# any subnet id where you want your nodes running'
    tags:
      myTag: 'MyValue # Use this tags to annotate nodepool resources. Optional'
  tags:
    myTag: 'MyValue # Use this tags to annotate all resources. Optional'
  auth:
    additionalAccounts:
    - "777777777777"
    - "88888888888"
    - '# Additional AWS account numbers to add to the aws-auth configmap'
    users:
    - username: 'user1 # Any username'
      groups:
      - system:masters
      - '# Any K8S Group'
      userarn: 'arn:user:7777777... # The user ARN'
    roles:
    - username: 'user1 # Any username'
      groups:
      - system:masters
      - '# Any K8S Group'
      rolearn: 'arn:role:7777777... # The role ARN'
executor:
  path: '/your/terraform/binary # Set this value if you need to use an already installed
    terraform binary'
  version: '0.12.29 # Set this value to download a specific terraform version. If
    version and path is not specified, latest terraform version will be downloaded'
  state:
    backend: 'local # Specify your backend configuration. local is the default one.
      https://www.terraform.io/docs/configuration/backend.html'
    config:
      path: 'workdir/terraform.state # Set the backend configuration parameter. this
        path attribute works with the local backend.'
provisioner: eks
```

Thanks!

angelbarrera92:
> Besides all the things already highlighted by @nutellinoit and @lzecca78, I would like to add the following:
> 
>     * avoid using a shared `sighup` user for the bastion (even though with multiple SSH keys)

This is something already implemented in the furyagent ssh user feature.
This feature creates a robot account (with is configurable) then adds ssh public keys from a welknown provider. Currently GitHub is being used.

> 
>     * in the bootstrap phase, add an option to skip creation of the VPC and use a customer-provided one

This is something that has to be enable first in the provisioner. It could makes sense.

>     * in the bootstrap phase, add an option to tweak the number of bastions create (which could also be 0)

Again, this is something that has to be enable first in the provisioner. It could makes sense.

> 
> I've mixed feelings about dumping the terraform code and allowing to fiddle with it, I think the aim of this work is to create an opinionated way to create Kubernetes clusters. Everything else should be handled with a separate Terraform project with a separate state.

The idea is to let use show the code, but the `furyctl` tool is always overriding the terraform code to avoid missconfigurations (intended or not). 

> Also, what about adding an `apiVersion` to the YAML files? I agree that right now is just overkill but in the future might well be put to good use.

Ye... I had it in the first iteration, but i don't know if it makes sense currently

angelbarrera92:
I've addressed almost all the points. At least those that I consider relevant for the MVP release.
Tomorrow I'll update the docs, meanwhile, I suggest to re-review the project :)

If I miss something relevant, ping me without any doubt

angelbarrera92:
I am adding this cvi doc to continue with the furyctl self-provisioning feature to add GKE / Google Cloud provisioners.

Issue #3, opened 2019-01-15
bash/zsh completion missing


gitirabassi:
It would be cool to have bash/zsh completion

nandajavarma:
This feat has been added in #42. Closing this issue.

PR #42, opened 2021-05-31
Adds (ba|z|fi)sh autocomplete for furyctl

nandajavarma:
This MR addresses issue #3. The cobra generator was used to auto-generate the completion scripts. The configuration to add the completion scripts to the shell configurations will be available on running the command `furyctl completion -h` as follows:
```bash
To load furyctl completions:

Bash:

  $ source <(furyctl completion bash)

  # To load completions for each session, execute once:
  # Linux:
  $ furyctl completion bash > /etc/bash_completion.d/furyctl
  # macOS:
  $ furyctl completion bash > /usr/local/etc/bash_completion.d/furyctl

Zsh:

  # If shell completion is not already enabled in your environment,
  # you will need to enable it.  You can execute the following once:

  $ echo "autoload -U compinit; compinit" >> ~/.zshrc

  # To load completions for each session, execute once:
  $ furyctl completion zsh > "${fpath[1]}/_furyctl"

  # You will need to start a new shell for this setup to take effect.

fish:

  $ furyctl completion fish | source

  # To load completions for each session, execute once:
  $ furyctl completion fish > ~/.config/fish/completions/furyctl.fish

PowerShell:

  PS> furyctl completion powershell | Out-String | Invoke-Expression

  # To load completions for every new session, run:
  PS> furyctl completion powershell > furyctl.ps1
  # and source this file from your PowerShell profile.

Usage:
  furyctl completion [bash|zsh|fish|powershell]

Flags:
  -h, --help   help for completion

Global Flags:
      --debug     Enables furyctl debug output
  -d, --disable   Disable analytics
```

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -1,18 +1,7 @@
-/*
-Copyright © 2021 NAME HERE <EMAIL ADDRESS>
+// Copyright (c) 2021 SIGHUP s.r.l All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
 package cmd
 
 import (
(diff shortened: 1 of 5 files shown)


cmd/completion.go
```diff
@@ -0,0 +1,85 @@
+/*
```
angelbarrera92, line 1: We should use our own license headers. To fix this:
- Remove the Apache header
- Install `addLicense` by running `go get -u github.com/google/addlicense`
- Ensure the License header on every file: `addlicense -c "SIGHUP s.r.l" -v -l bsd cmd/`

angelbarrera92 (approved):
Once you solve the "license" header problem, we can deliver it!

Amazing job @nandajavarma :)

cmd/completion.go
nandajavarma: Whoops! Thanks for the review @angelbarrera92 ! Just pushed a commit fixing this. :)

PR #43, opened 2021-06-01
handling spotInstancePrice value

lzecca78:
Hi everyone 👋 ,
is still young this PR, but probably it already contains the main changes for implementing the spot instance feature in the furyctl self provisioner for EKS. 
Still missing, as @angelbarrera92  already knows, the common refactoring for putting this feature on top of the 3 cloud providers that we want to manage (the aim is to have an agnostic configuration that is then implemented in the 3 cloud provider, in order to give to the user the availability to handle in the same way this capability no matter what is the cloud provider).
Is a step ahead in that direction, but before extract this refactoring, is better to have a working implementation 👍 
Any feedbacks are welcome! 🙏


lnovara (approved):
LGTM :shipit:

nutellinoit (changes requested):
Opened a discussion here https://github.com/sighupio/fury-eks-installer/pull/19 

I would like something to enable spot (if possible), true or false, without setting the price

PR #46, opened 2021-07-05
Hotfix ami owners eks

angelbarrera92:
This address the problem you found in eu-south region


lbottalico:
Hi @angelbarrera92 

seems it works, thank you!

PR #49, opened 2021-08-06
handling spotInstancePrice value

lzecca78:
Reopened this PR for some minor fix to be done @emasacco


automated-tests/integration/gcp-gke/cluster.yml
```diff
@@ -21,3 +21,9 @@ spec:
     maxSize: 1
     volumeSize: 50
     instanceType: n1-standard-1
+  - name: my-spot-node-pool
+    minSize: 1
+    maxSize: 1
+    volumeSize: 50
+    spotInstance: true
```
angelbarrera92, line 28: AFAIK, this attribute is wrong. You named it different on the `gke` installer

internal/cluster/configuration/gke.go
```diff
@@ -29,6 +29,7 @@ type GKENodePool struct {
 	MinSize                 int                 `yaml:"minSize"`
 	MaxSize                 int                 `yaml:"maxSize"`
 	InstanceType            string              `yaml:"instanceType"`
+	SpotInstance            bool                `yaml:"spotInstance"`
```
angelbarrera92, line 32: Same here

internal/cluster/provisioners/gke/provisioner.go
```diff
@@ -128,6 +129,8 @@ func (e GKE) createVarFile() (err error) {
 			buffer.WriteString(fmt.Sprintf("instance_type = \"%v\"\n", np.InstanceType))
 			buffer.WriteString(fmt.Sprintf("max_pods = %v\n", np.MaxPods))
 			buffer.WriteString(fmt.Sprintf("volume_size = %v\n", np.VolumeSize))
+			buffer.WriteString(fmt.Sprintf("preemptible = %v\n", np.SpotInstance))
```
angelbarrera92, line 132: Be consistent...

lzecca78: @angelbarrera92, you asked for a `soft interface` many times, saying that was the added value that SIGHUP put into. I just tried to comply with your requests keeping the same value in both the cloud providers. In the terraform of GKE, I have to maintain the ubiquitous language that google chooses for the name of that kind of instance.
This is why I opted for keeping in the yaml the same `spotInstance` name for both providers (because is a level of indirection) while in the underneath implementation of the terraform I kept the domain right name for both providers.

angelbarrera92: Sure!
What I don't agree with is using different attribute names on the `terraform modules`. It triggers exactly these confusions/doubts...

lzecca78: the terraform code is on the domain side of the cloud provider, so for respecting the domain ubiquitous language, I must put the exact name that the cloud provider involves.

angelbarrera92: You may be right, but if you notice, there are already attributes with standard (our standard) names in terraform modules... like:

- subnetworks: on AWS are subnet-ids while on GKE is a different concept closer to `availability zones`.
- additional_firewall_rules: On AWS are security groups while on GKE are firewall rules.

Then, when we need some specific attributes on gke (for example) we decided to go with the same attribute name from the upstream module plus the `gke_` prefix.

Example: https://github.com/sighupio/furyctl/blob/master/data/provisioners/cluster/gke/variables.tf#L76-L80

TL;DR: What I want is consistency, that's all.

automated-tests/integration/gcp-gke/cluster.yml
nandajavarma: made the variable name consistently `spot_instance`

data/provisioners/bootstrap/gcp/main.tf
```diff
@@ -17,7 +17,7 @@ terraform {
 
 
 module "vpc-and-vpn" {
-  source = "github.com/sighupio/fury-gke-installer//modules/vpc-and-vpn?ref=v1.7.0"
+  source = "github.com/sighupio/fury-gke-installer//modules/vpc-and-vpn?ref=v1.7.1"
```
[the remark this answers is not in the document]
nandajavarma: I have updated this for now, I will add a small hook to update it automatically

internal/cluster/configuration/gke.go
nandajavarma: fixed!

nutellinoit:
To be re-tested

.drone.yml
```diff
@@ -89,84 +92,84 @@ steps:
...
-#      - echo " Our vsphere environment is not so stable"
-#      - echo " The destroy phase fails randomly because of disconnection events between vsphere nodes and the API"
-#      - echo " Ignore errors on this step"
-#      - bats -t ./automated-tests/e2e-tests/vsphere/tests-destroy.sh
+
+  - &e2e
+    name: e2e-gcp
+    image: quay.io/sighup/e2e-furyctl:v1.2.1_v0.2.2_v1.20.1_20.04
+    pull: always
+    privileged: true # Required to connect to the VPN
```
ralgozino, line 100: If the pipeline runs on drone master, we don't need this anymore
```suggestion

```
This was needed when we had drone hosted on the cloud instead of our vSphere env

ralgozino (approved):
just a minor change to the drone.yaml file. Feel free to merge afterwards

.drone.yml
nutellinoit: We need the parameter for the VPN connection on EKS and GKE environments

PR #57, opened 2021-11-19
Log the error in the case of wrong tag for modules

nandajavarma:


cmd/download.go
```diff
@@ -78,33 +78,38 @@ func get(src, dest string, mode getter.ClientMode, cleanGitFolder bool) error {
...
 	}
 
 	humanReadableDownloadLog(src, dest)
-	_ = client.Get()
+
+	err = client.Get()
+	if err != nil {
+		return err
+	}
+
```
ralgozino, line 106: so, the bug was that we were failing when the wrong tag/branch/version was specified but we were failing silently? 🙃

ralgozino (approved):
lgtm

cmd/download.go
nandajavarma: Yes, basically! :D

PR #61, opened 2022-02-16
[BUGFIX] don't delete Furyfile if wrong KFD --version tag is supplied

nutellinoit:
In the event of a wrong supplied tag, furyctl removes the Furyfile.yml.

### DIFF
--- a/cmd/download.go
+++ b/cmd/download.go
@@ -74,6 +74,8 @@ func get(src, dest string, mode getter.ClientMode, cleanGitFolder bool) error {
 
 	logrus.Debugf("complete url downloading: %s -> %s", src, dest)
 
+	var tempDest = dest + ".tmp"
+
 	pwd, err := os.Getwd()
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


cmd/download.go
```diff
@@ -74,34 +74,44 @@ func get(src, dest string, mode getter.ClientMode, cleanGitFolder bool) error {
...
 
 	client := &getter.Client{
 		Src:  src,
-		Dst:  dest,
+		Dst:  tempDest,
 		Pwd:  pwd,
 		Mode: mode,
 	}
 
 	logrus.Debugf("let's get %s -> %s", src, dest)
```
ralgozino, line 91: this debug log will be misleading, because it will actually be downloaded to `tmpDest` instead of `dest`.

cmd/download.go
```diff
@@ -74,34 +74,44 @@ func get(src, dest string, mode getter.ClientMode, cleanGitFolder bool) error {
...
-		err = removeDir(dest)
-		if err != nil {
-			logrus.Error(err)
-			return err
-		}
-	}
-
 	humanReadableDownloadLog(src, dest)
 
 	err = client.Get()
```
ralgozino, line 95: should we check if `tmpDest` exists and clean it up also? It might exists because some previous operation that didn't go as expected for example.

cmd/download.go
```diff
@@ -145,3 +155,11 @@ func removeDir(dir string) error {
 	}
 	return nil
 }
+
+func renameDir(src string, dest string) error {
+	err := os.Rename(src, dest)
```
ralgozino, line 160: maybe would be nice to check here if `src` and `dest` are folders here before renaming?

cmd/download.go
```diff
@@ -74,34 +74,44 @@ func get(src, dest string, mode getter.ClientMode, cleanGitFolder bool) error {
...
-		}
-	}
-
 	humanReadableDownloadLog(src, dest)
 
 	err = client.Get()
 	if err != nil {
+		_ = removeDir(tempDest)
 		return err
+	}else{
```
ralgozino, line 99: I'm pretty sure this line is breaking the linter. Should be `} else {`. Please run `go fmt` on the file :)

ralgozino (changes requested):
looks good, I left some minor comments

cmd/download.go
nutellinoit: Will do 👍

cmd/download.go
nutellinoit: Yes, we should 👍

cmd/download.go
nutellinoit: holy 😄

nutellinoit: Done, i'm adding the target lint to the build command, since the linting is automated

ralgozino (approved):
lgtm!
it would be nice to have a golang dev POV so we can merge.
pinging @lzecca78 💪

lzecca78 (approved):
LGTM

PR #62, opened 2022-02-16
[REFACTORING] Move init and vendor (renamed as download) under distribution command

nutellinoit:
Hi team, this change is to enhance and make the management of the KFD easy and usable with the same interface 
<command> <subcommand> of all the other furyctl commands.

1) `init` command has been moved under `distribution` command, and by default, the kustomize.yaml file is not downloaded, there is `--kustomize` flag to download it
2) `vendor` command has been renamed in `download`, this way what furyctl does is easily readable

tell me what you think, obviously this is a breaking change


cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) 2020 SIGHUP s.r.l All rights reserved.
```
ralgozino, line 1: 2022 
> Captain: what a year, huh'?
> Ttintin: Captain, it's 2022

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+			viper.SetConfigType("yml")
+			viper.AddConfigPath(".")
+			viper.SetConfigName(configFile)
+			config := new(Furyconf)
+			if err := viper.ReadInConfig(); err != nil {
+				logrus.Fatalf("Error reading config file, %s", err)
+			}
+			err := viper.Unmarshal(config)
+			if err != nil {
+				logrus.Fatalf("unable to decode into struct, %v", err)
```
ralgozino, line 80: can we improve this error message to make it more user-friendly?

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+
+			err = config.Validate()
+			if err != nil {
+				logrus.WithError(err).Error("ERROR VALIDATING")
```
ralgozino, line 85: I'm not sure about what is the end result (I'm not familiar with WithError and the error that one could get in this case) is "ERROR VALIDATING" and the `WithError` enough to understand what is happening?

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+
+			if err != nil {
+				//logrus.Errorln("ERROR PARSING: ", err)
+				logrus.WithError(err).Error("ERROR PARSING")
+
+			}
+
+			err = download(list)
+			if err != nil {
+				//logrus.Errorln("ERROR DOWNLOADING: ", err)
```
ralgozino, line 98: delete the comment

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+				logrus.WithError(err).Error("ERROR DOWNLOADING")
```
ralgozino, line 99: same as with `ERROR VALIDATING`

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+	return err
+}
+
+func init() {
+
+	initCmd.PersistentFlags().StringVarP(&distributionVersion, "version", "v","", "Specify the Kubernetes Fury Distribution version")
+	initCmd.PersistentFlags().BoolVar(&initKustomize, "kustomize", false,"Initialize kustomize.yaml file")
+
+	downloadCmd.PersistentFlags().BoolVarP(&parallel, "parallel", "p", true, "if true enables parallel downloads")
+	downloadCmd.PersistentFlags().BoolVarP(&https, "https", "H", false, "if true downloads using https instead of ssh")
```
ralgozino, line 128: same as before, these are help messages not code comments :)

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+	downloadCmd.PersistentFlags().StringVarP(&prefix, "prefix", "P", "", "Add filtering on download with prefix, to reduce update scope")
```
ralgozino, line 129: I don't understand what this flag does from the help message

ralgozino (changes requested):
Hi Samu!

I'm totally in favour of structuring better the commands 💪 

About the `vendor` -> `download` change, it was previously `install` and I proposed to change it to `download` or `vendor` and the latter was chosen for some reason that I can't remember. I'm fine with both of them.

For the `kustomization.yaml` file download could you elaborate a little more on the rationale please? Thanks!

let's keep these changes in queue tough, we have some more breaking stuff in pipeline that we could release together.

cmd/distribution.go
nutellinoit: Yes, but it's something I did not write, i inherited from the refactoring

cmd/distribution.go
nutellinoit: Yes, but it's something I did not write, i inherited from the refactoring

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+
+			err = config.Validate()
+			if err != nil {
+				logrus.WithError(err).Error("ERROR VALIDATING")
+			}
+
+			list, err := config.Parse(prefix)
+
+			if err != nil {
+				//logrus.Errorln("ERROR PARSING: ", err)
```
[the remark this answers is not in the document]
nutellinoit: Yes, but it's something I did not write, i inherited from the refactoring

cmd/distribution.go
nutellinoit: Yes, but it's something I did not write, i inherited from the refactoring

cmd/distribution.go
```diff
@@ -0,0 +1,135 @@
...
+				logrus.WithError(err).Error("ERROR PARSING")
```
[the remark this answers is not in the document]
nutellinoit: Yes, but it's something I did not write, i inherited from the refactoring

cmd/distribution.go
nutellinoit: Ok, same as above, inherited :D

cmd/distribution.go
nutellinoit: Ok, same as above, inherited :D

cmd/distribution.go
nutellinoit: I'll open another PR for the cosmetic issues

nutellinoit:
Opened #65 PR to address out of scope changes

ralgozino:
I was forgetting one very important part, the docs need to be updated to reflect this change as part of this PR.

We'll need to track somewhere else also to upgrade all the quick start guides and modules docs that have instructions using the current commands structure. This one is out of the scope of this PR.

nutellinoit:
> I'm not seeing why would one use `init` at T-1, T-2, etc. at all... could illuminate me, please?

Right now it's the only automated way to download the updated Furyfile.yml. If you already have a cluster you will have a Kustomize project already in place, I personally don't find it useful that a command that manages the distribution does something bonus not strictly required. We can switch the logic tho, or remove this change if we considered to be out of scope

> > Absolutely, if the implementation is ok I can start with the docs. Maybe we can create a new dev branch on the repository to merge all the changes, this way we can see with a single PR to master/main what is going to be released.

Sure 👍

nutellinoit:
Renamed `init` command to `template` command, refactored `--kustomize` flag to `--skip-kustomize`

lzecca78 (approved):
On top of @ralgozino already suggested, i don't see any issue with this, LGTM

PR #63, opened 2022-02-16
[FEATURE] Add update subcommand to distribution command

nutellinoit:
Hi team,

this PR is dependent to the PR #62 

I implemented an experimental command to update in place the `Furyfile.yml` file based on a new KFD version. What it does is a simple merge with the upstream Furyfile.yml with the current one

I think it can be useful 😃 tell me what you think!


lzecca78 (approved):
LGTM

ralgozino (approved):
I see some stuff that we've already discussed and fixed on #62, maybe this branch needs a rebase?

Feature-wise I think this is a useful feature for the delivery team, so let's add it :) 

Let's see how the docs come from #62 maybe we can add something about this feature.

PR #81, opened 2022-07-05
Feature/template engine

Al-Pragliola:
This PR aims to add a template command to create a cluster's folder and file structure based on a distribution template
 
Features
- [x] support to env:// and file:// to get content from environment variables or files at runtime
- [x] --dry-run flag to log into a file all keys from template files that are missing values
- [x] merge furyctl.yaml into distribution.yaml

Notes

source, target, furyctl.yaml, distribution.yaml's paths are hardcoded


lzecca78 (changes requested):
for what i can see in the very big picture, except my 2 cosmetic reviews, SBAM

nutellinoit:
I think that `master` is not the correct branch this PR should be aimed to. 
Since these are new implementation in the new `furyctl` stream, it's better to aim the PR to a new development branch (i don't know yet how it will be named, maybe `release-v0.10`. 
Correct me if i'm wrong @ralgozino

ralgozino:
That's correct @nutellinoit , the target branch for this feature should be the branch where we'll develop all the next gen furyctl, and we have agreed to use `release-v0.10` as branch name

ralgozino (changes requested):
as per the previous comment

ralgozino (changes requested):
Hi Ale!
I built the branch locally and tested the `template` command. I've found that:
- If the `distribution.yaml` and/or the `furyctl.yaml` files are empty, furcytl panics with and traceback that is not easy to understand. We should handle that error.
- If the `source` folder does not exists, furyctl panics with a traceback that is not easy to understand. We should handle that error.
- If there's an error in the template definition, for example accessing a key that does not exist you get the error printed twice, seems that one of the errors comes from cobra (because the usage is printed also) and the other is the actual error from the logger. (Maybe this is happening with other commands and is not an issue of the template command, I did not check). I believe the error from cobra should not be triggered.
- Speaking of keys that do not exists, there's a change in the engine behaviour from the POC. In the POC I could do this: `{{if ne .xxx.yyy false}}<do something>{{end}}` even if `yyy` did not exists as a key, but with this version the same thing fails with a `map has no entry for key "enabled"` error. Is this some side effect of another thing? The behaviour like in the POC is desirable and useful.

Thanks!

internal/io/fs.go
```diff
@@ -0,0 +1,94 @@
...
+		err := filepath.Walk(target, func(path string, info os.FileInfo, err error) error {
+			return fmt.Errorf("the target directory is not empty: %s", path)
+		})
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
Al-Pragliola: added a --no-overwrite flag to stop if the user doesn't want to clear the target dir before generating a new template

ralgozino (approved):
feature wise, lgtm 🚀

cmd/template.go
```diff
@@ -0,0 +1,145 @@
...
+				logrus.Errorf("%+v", err)
+				return nil
+			}
+
+			dss, _ := cmd.Flags().GetStringSlice("datasource")
+
+			if len(dss) > 0 {
+				if templateModel.Config.Include == nil {
+					templateModel.Config.Include = make(map[string]string)
+				}
```
omissis, line 115: given you call NewTemplateModel a few lines above, you can move this logic within that function so that Include is ready for use and client code doesn't have to initialize it separately

internal/io/fs.go
```diff
@@ -0,0 +1,91 @@
...
+	if err != nil {
+		return err
+	}
+
+	_, err = b.WriteTo(destination)
+	if err != nil {
+		return err
+	}
+
+	defer destination.Close()
```
omissis, line 37: deferring here is of little value, either call it directly or (better) move it before `b.WriteTo(destination)`

internal/io/fs.go
```diff
@@ -0,0 +1,91 @@
...
```
omissis, line 60: move it before `b.WriteTo(destination)`

internal/io/fs.go
```diff
@@ -0,0 +1,91 @@
...
+	return nil
+}
+
+func CopyBufferToFile(b bytes.Buffer, source, target string) error {
+	if strings.TrimSpace(b.String()) == "" {
+		fmt.Printf("%s --> resulted in an empty file (%d bytes). Skipping.\n", source, b.Len())
+		return nil
+	}
+
+	fmt.Printf("%s --> %s\n", source, target)
```
omissis, line 48: remove, unless you meant to use the logger

internal/io/fs.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	defer destination.Close()
+
+	return nil
+}
+
+func CopyBufferToFile(b bytes.Buffer, source, target string) error {
+	if strings.TrimSpace(b.String()) == "" {
+		fmt.Printf("%s --> resulted in an empty file (%d bytes). Skipping.\n", source, b.Len())
```
omissis, line 44: it looks like this should use the logger instead of fmt

internal/io/fs.go
```diff
@@ -0,0 +1,91 @@
...
+	"strings"
+)
+
+func CheckDirIsEmpty(target string) error {
+	if _, err := os.Stat(target); os.IsNotExist(err) {
+		return nil
+	}
+
+	return filepath.Walk(target, func(path string, info os.FileInfo, err error) error {
+		return fmt.Errorf("the target directory is not empty: %s", path)
```
omissis, line 22: you might want to embed `err` here, to provide more information when the program exits

internal/io/fs.go
```diff
@@ -0,0 +1,91 @@
...
+
+	destination, err := os.Create(dst)
+	if err != nil {
+		return 0, err
+	}
+
+	defer destination.Close()
+
+	nBytes, err := io.Copy(destination, source)
+	return nBytes, err
```
omissis, line 90: nitpick
```suggestion
	return io.Copy(destination, source)
```

internal/merge/model.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func NewDefaultModel(content map[string]any, path string) *DefaultModel {
+	return &DefaultModel{
+		content: content,
+		path:    path,
+	}
+}
+func (b *DefaultModel) Content() map[string]any {
+	return (*b).content
```
omissis, line 31: won't `b.content` work?

internal/merge/model.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (b *DefaultModel) Path() string {
+	return (*b).path
```
omissis, line 35: won't `b.path` work?

internal/merge/model.go
```diff
@@ -0,0 +1,78 @@
...
+	Content() map[string]any
+	Path() string
+}
+
+type DefaultModel struct {
+	content map[string]any
+	path    string
+}
+
+func NewDefaultModel(content map[string]any, path string) *DefaultModel {
```
omissis, line 24: we can probably implement this code using generics, thus improving type safety

internal/merge/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"github.com/sighupio/furyctl/internal/merge"
+)
+
+func TestNewDefaultModel(t *testing.T) {
+	content := map[string]any{
+		"data": map[string]any{
+			"meta": map[string]string{
+				"name": "pippo",
```
omissis, line 19: nitpick
```suggestion
				"name": "example",
```

internal/merge/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+	assert.NotEmpty(t, model)
+	assert.Equal(t, content, model.Content())
+	assert.Equal(t, path, model.Path())
+}
+
+func TestDefaultModel_Content(t *testing.T) {
+	content := map[string]any{
+		"data": map[string]any{
+			"meta": map[string]string{
+				"name": "pippo",
```
omissis, line 40: nitpick
```suggestion
				"name": "example",
```

internal/merge/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+	model := merge.NewDefaultModel(content, path)
+
+	assert.Equal(t, content, model.Content())
+}
+
+func TestDefaultModel_Path(t *testing.T) {
+	content := map[string]any{
+		"data": map[string]any{
+			"meta": map[string]string{
+				"name": "pippo",
```
omissis, line 59: nitpick
```suggestion
				"name": "example",
```

internal/merge/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+	model := merge.NewDefaultModel(content, path)
+
+	assert.Equal(t, path, model.Path())
+}
+
+func TestDefaultModel_Get(t *testing.T) {
+	content := map[string]any{
+		"data": map[string]any{
+			"meta": map[string]string{
+				"name": "pippo",
```
omissis, line 78:
```suggestion
				"name": "example",
```

internal/merge/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	assert.NoError(t, err)
+	assert.Equal(t, expectedRes, res)
+}
+
+func TestDefaultModel_Walk(t *testing.T) {
+	content := map[string]any{
+		"data": map[string]any{
+			"meta": map[string]string{
+				"name": "pippo",
```
omissis, line 104:
```suggestion
				"name": "example",
```

internal/merge/model_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	target := map[string]any{
+		"testString":    "lorem ipsum",
+		"testNewString": "lorem ipsum new",
+	}
+
+	expectedRes := map[string]any{
+		"data": map[string]any{
+			"meta": map[string]string{
+				"name": "pippo",
```
omissis, line 120:
```suggestion
				"name": "example",
```

internal/template/generator.go
```diff
@@ -0,0 +1,148 @@
...
+func (g *generator) WriteMissingKeysToFile(
+	missingKeys []string,
+	tmplPath,
+	outputPath string,
+) error {
+	if len(missingKeys) == 0 {
+		return nil
+	}
+
+	fmt.Printf(
```
omissis, line 113: you might want to use the logger instead of fmt here

omissis (changes requested):
The command help does not tell the user what's needed or how to properly run the templating.
Besides what's written inline, please also consider:
- renaming the`--disable` flag to `--disable-analytics`
- add better description on how the command should be used
- add at least one functional test exercising the command from the cli, as I couldn't find a way to make it work even when using the files provided in the POC repository

cmd/template.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/sighupio/furyctl/internal/template"
+)
+
+var (
+	tDryRun bool
+
+	TemplateCmd = &cobra.Command{
+		Use:   "template",
+		Short: "This is a POC for furyctl's Template Engine in go.",
+		Long:  `This is a POC for furyctl's Template Engine in go.`,
```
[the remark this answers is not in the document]
ralgozino: @Al-Pragliola asked me about this before opening the PR and I said it was OK. Not his fault.

Here's a proposal:

```suggestion
		Short: "Renders the distribution's manifests from a template and a configuration file",
		Long:  `Generates a folder with the Kustomization project for deploying Kubernetes Fury Distribution into a cluster.
The generated folder will be created starting from a provided template and the parameters set in a configuration file that is merged with default values.`,
```

internal/merge/model.go
Al-Pragliola: yup, we should invest some time to check

related task: [rm-2471](https://redmine.sighup-prod.sighup.io/issues/2471)

cmd/template.go
```diff
@@ -0,0 +1,145 @@
...
+	tDryRun      bool
+	tNoOverwrite bool
+
+	TemplateCmd = &cobra.Command{
+		Use:   "template",
+		Short: "This is a POC for furyctl's Template Engine in go.",
+		Long:  `This is a POC for furyctl's Template Engine in go.`,
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			//TODO: Hardcoded for now, we have to think a final strategy for them.
```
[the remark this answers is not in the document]
Al-Pragliola: linked to redmine task [rm-2470](https://redmine.sighup-prod.sighup.io/issues/2470)

automated-tests/integration/template-engine/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
omissis, line 1:
```suggestion
target

```

automated-tests/integration/template-engine/test-data/simple-dry-run/source/file.txt.tpl
```diff
@@ -0,0 +1 @@
+{{.test.hello}}
```
omissis, line 1:
```suggestion
{{.test.hello}}

```

automated-tests/integration/template-engine/test-data/simple/source/file.txt.tpl
```diff
@@ -0,0 +1 @@
+{{.test.hello}}
```
omissis, line 1:
```suggestion
{{.test.hello}}

```

omissis (changes requested):
It would be nice to have an integration test that covers realistic distribution and furyctl files, so to offer a real-ish-life example and not just a "trivial" one, perhaps @ralgozino can also help with that, so that I can test it, too? Besides this and some missing NLs, I'm good.

omissis (approved):

ralgozino (approved):
lgtm

PR #82, opened 2022-07-07
fix: provisioner flag is case sensitive

luigibarbato:
Currently, the provisioner flag is case sentive. This PR aims to solve this problem, lowerizing the provisioner flag value.


ralgozino:
Hi @luigibarbato , I think that you meant that the flag IS case sensitive, right? 
(and `currently` instead of `actually`)

can we point this PR to `master` instead of `fury-next` please? The `fury-next` branch will be removed.

Thanks!

ralgozino (changes requested):
as per the previous comment

luigibarbato:
> Hi @luigibarbato , I think that you meant that the flag IS case sensitive, right? (and `currently` instead of `actually`)
> 
> can we point this PR to `master` instead of `fury-next` please? The `fury-next` branch will be removed.
> 
> Thanks!

Oh! You're right! I fixed it :)

One question on PR's pointer: Currently, the PR codebase is based on furyctl-next code and the PR branch is aligned with the furyctl-next branch. We would wait for furyctl-next merge so the PR automatically points to master or we would rebase the actual branch with the master branch?

ralgozino:
> One question on PR's pointer: Currently, the PR codebase is based on furyctl-next code and the PR branch is aligned with the furyctl-next branch. We would wait for furyctl-next merge so the PR automatically points to master or we would rebase the actual branch with the master branch?

Hi Luigi, we should rebase and make the PR to `master`, the content in `furyctl-next` is no longer valid and we'll delete it.

omissis:
closed in favor of #85

PR #83, opened 2022-07-07
fix: added valid options for provisioners

Al-Pragliola:
This PR aims to add the list of supported provisioners to cluster/bootstrap template cmd

### DIFF
--- a/cmd/bootstrap.go
+++ b/cmd/bootstrap.go
@@ -8,7 +8,6 @@ import (
 	"bufio"
 	"errors"
 	"fmt"
-	"github.com/sighupio/furyctl/internal/provisioners"
 	"os"
 	"os/signal"
 	"strings"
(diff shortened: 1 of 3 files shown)


omissis (commented):
While this is a small pr, I find it a bit over-engineered. Adding the list of providers straight in the help text string would do the same without introducing extra complexity, i.e.:
```
"Bootstrap provisioner, valid options are: aws, gcp"
```

ralgozino:
Hi @Al-Pragliola 
I kind of agree with @omissis on this one. If we can't get/detect the supported list in a smarter way and we'll need to maintain the variable in the provisioners.go file, it would be simpler to just have it in the help text.

Also, can we point this PR to `master` instead of `fury-next` please? The `fury-next` branch will be removed.

Thanks!

ralgozino (changes requested):
as per the previous comment

omissis (approved):
SBAM

ralgozino (approved):

PR #85, opened 2022-07-25
fix: provisioner flag is case sensitive

luigibarbato:
Currently, the provisioner flag is case sentive. This PR aims to solve this problem, lowerizing the provisioner flag value.


omissis (approved):

ralgozino (approved):
Applied the same fix to the `furyctl cluster template --provisioner` command
lgtm!

PR #95, opened 2022-09-20
feat(template): load _helpers.tpl together with the template

ralgozino:
load a `_helpers.tpl` file if exists and load it together with the temlpate to be rendered, allowing to define helper functions in the template logic.

This allows for example to have common logic shared between template files, for example, e function to get the ingress class.

Please feel free to modify the code as you wish

### DIFF
--- a/internal/template/generator.go
+++ b/internal/template/generator.go
@@ -6,7 +6,9 @@ package template
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strings"
 	"text/template"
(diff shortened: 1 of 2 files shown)


internal/template/generator.go
```diff
@@ -40,8 +42,18 @@ func NewGenerator(
...
+		return template.Must(
+			template.New(filepath.Base(g.source)).Funcs(g.funcMap.FuncMap).ParseFiles(g.source, helpersPath))
+	} else if errors.Is(err, os.ErrNotExist) {
+		logrus.Warnf("template helpers file '%s' not found\n", helpersPath)
+		return template.Must(
+			template.New(filepath.Base(g.source)).Funcs(g.funcMap.FuncMap).ParseFiles(g.source))
+	} else {
+		logrus.Fatalf("got unexpected error when checking if helper file exists at path %s: %s\n", helpersPath, err)
+		panic(err)
+	}
```
omissis, line 56: Please refrain from panic-ing in library code: [return descriptive errors instead](https://go.dev/doc/effective_go#errors).
In this case Fatalf already exits, so this panic call will never get called and can be removed.
More info about Go "Zen" can be found here: https://go-proverbs.github.io

omissis (changes requested):

internal/template/generator.go
ralgozino: the `Fatalf` exiting was my understanding but the VSCode linter was still asking for a return. I've deleted the `panic()` line.

Thanks for the links 🙏

omissis (approved):
LGTM

PR #100, opened 2022-09-29
feature/ginkgo e2e tests

omissis:
- feat: replace bats tests with ginkgo ones, add workdir flag
- feat: add test for download deps, refactor e2e tests
- feat: tag all test files with build tag, separate integration test data from e2e's
- feat: add test for create config command


Al-Pragliola (approved):
LGTM

PR #101, opened 2022-09-29
feature/further code cleanup

omissis:
- chore: remove old code
- chore: remove packr go library
- chore: remove old deps from manifests
- chore: bring back remaining logic from internal/app to cmd
- chore: update deps


Al-Pragliola (approved):
SBAM

PR #102, opened 2022-09-30
feat: add new release's update check

luigibarbato:


cmd/root.go
```diff
@@ -66,6 +73,17 @@ Furyctl is a simple CLI tool to:
 				analytics.Version(versions["version"])
 				analytics.Disable(cobrax.Flag[bool](cmd, "disable-analytics").(bool))
 			},
+			PersistentPostRun: func(cmd *cobra.Command, _ []string) {
+				// Show update message if available at the end of the command
+				select {
+				case release := <-r:
+					if release.Version != versions["version"] {
```
Al-Pragliola, line 80: nitpick because it won't be happening in a normal use case but for example with versions: 0.10.0 (branch) and 0.8.0 (GitHub latest tag) it still says:
```
INFO[0000] New furyctl version available: 0.10.0 => v0.8.0
```

cmd/root.go
```diff
@@ -85,3 +103,19 @@ Furyctl is a simple CLI tool to:
 
 	return rootCmd
 }
+
+func checkUpdates(version string, rc chan app.Release, e chan error) {
+	if version == "unknown" {
+		return
```
Al-Pragliola, line 109: in this case, the program hangs forever

internal/app/update.go
```diff
@@ -0,0 +1,39 @@
...
+type Release struct {
+	URL     string `json:"html_url"`
+	Version string `json:"name"`
+}
+
+func GetLatestRelease() (Release, error) {
+	var release Release
+
+	ctx := context.Background()
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "https://api.github.com/repos/sighupio/furyctl/releases/latest", nil)
```
Al-Pragliola, line 22: it would be nice to have a timeout in the context so as to not hang the command when the user is offline

main.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	version   = "unknown"
+	version   = "0.2.0"
```
Al-Pragliola, line 24: forgot to remove it?

Al-Pragliola (changes requested):

internal/app/update.go
```diff
@@ -0,0 +1,90 @@
...
+func ShouldUpdate(currentVersion, latestVersion string) bool {
+	// normalize the versions
+	nc := strings.TrimPrefix(currentVersion, "v")
+	nl := strings.TrimPrefix(latestVersion, "v")
+
+	return compareVersions(nc, nl)
+}
+
+// util func to compare two semantic versions, e.g: 0.8.0 vs 0.9.0
+func compareVersions(currentVersion, latestVersion string) bool {
```
Al-Pragliola, line 57: there's a semver package with a compare function in internal that you can check or add it there if needed

Al-Pragliola (changes requested):

omissis (approved):

PR #109, opened 2022-11-02
feat: add analytics tracker and events

luigibarbato:
This PR aims to implements the following analytics capabilities:
-  Version
-  Command
-  Architecture (amd, arm, etc.)
-  Operating System
-  Phase if available
-  Provider if available
-  Success / Failure status
-  KFD version if available
-  Organization if possible
-  Hostname / IP
-  Location


.gitignore
```diff
@@ -10,3 +10,5 @@ dist/
 .kubernetes
 .distribution
 test/data/e2e/**/target
+
+.vscode/launch.json
```
omissis, line 14: We should probably ignore the whole folder

cmd/completion.go
```diff
@@ -61,23 +70,27 @@ func NewCompletionCmd() *cobra.Command {
...
 				}
 			case "fish":
 				if err := cmd.Root().GenFishCompletion(os.Stdout, true); err != nil {
-					return fmt.Errorf("error generating fish completion: %w", err)
+					return ErrFishCompletion
 				}
 			case "powershell":
 				if err := cmd.Root().GenPowerShellCompletion(os.Stdout); err != nil {
-					return fmt.Errorf("error generating powershell completion: %w", err)
+					return ErrPowershellCompletion
```
omissis, line 85: Please do not hide `err` here, write `fmt.Errorf("%w: %v", ErrXyz, err) instead

cmd/create/cluster.go
```diff
@@ -141,8 +143,22 @@ func NewClusterCmd(version string) *cobra.Command {
...
 			}
 
+			cmdEvent = analytics.NewCommandEvent(cmd.Name(), "", 0, &analytics.ClusterDetails{
+				Phase:      phase,
+				Provider:   res.DistroManifest.Kubernetes.Eks.Version,
+				KFDVersion: res.DistroManifest.Version,
+			})
+
+			_, err = fmt.Println("cluster creation succeeded")
+			if err != nil {
```
omissis, line 153: nitpick
```suggestion
			if _, err = fmt.Println("cluster creation succeeded"); err != nil {
```

omissis (changes requested):

cmd/create/cluster.go
```diff
@@ -35,45 +35,74 @@ var (
...
 			}
 
 			// Init paths.
 			homeDir, err := os.UserHomeDir()
 			if err != nil {
-				return fmt.Errorf("error while getting user home directory: %w", err)
+				cmdEvent.AddErrorMessage(err)
+				tracker.Track(cmdEvent)
+
+				return fmt.Errorf("error while getting current working directory: %w", err)
```
omissis, line 105:
```suggestion
				return fmt.Errorf("error while getting user home directory: %w", err)
```

cmd/download/dependencies.go
```diff
@@ -27,35 +28,59 @@ var (
...
 				return ErrDistroLocationNotSet
 			}
 
 			homeDir, err := os.UserHomeDir()
 			if err != nil {
-				return fmt.Errorf("failed to get current user home directory: %w", err)
+				cmdEvent.AddErrorMessage(err)
+				tracker.Track(cmdEvent)
+
+				return fmt.Errorf("failed to get current working directory: %w", err)
```
omissis, line 68:
```suggestion
				return fmt.Errorf("error while getting user home directory: %w", err)
```

internal/analytics/tracker.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+}
+
+// Flush flushes the events queue, guaranteeing that all events are sent to mixpanel.
+// This method uses a timeout to send a GuardEvent to the event processor to close the process.
+func (a *Tracker) Flush() {
+	const timeout = time.Millisecond * 500
+
+	go func() {
+		time.Sleep(timeout)
```
omissis, line 81: We realized timeout isn't needed, remove it

internal/x/cobra/cmd.go
```diff
@@ -0,0 +1,20 @@
...
+package cobrax
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+// GetFullname returns the hierarchy of the command and its parents. For example: "<command> <subcommand>...".
+func GetFullname(c *cobra.Command) string {
```
omissis, line 14: nice one. it could also be nice to expand on this in the future with something like a `superhelp` command that prints all the commands and subcommands in one go.

main.go
```diff
@@ -41,11 +44,48 @@ func main() {
...
+	h, err := os.Hostname()
+	if err != nil {
+		logrus.Debug(err)
+
+		h = "unknown"
 	}
+
+	t := os.Getenv("FURYCTL_MIXPANEL_TOKEN")
+	if t == "" {
+		panic("FURYCTL_MIXPANEL_TOKEN environment variable not set")
```
omissis, line 63: in this case is not terrible but in general do not use panic unless you really need it.
go for a log + os.Exit(1) instead

main.go
```diff
@@ -41,11 +44,48 @@ func main() {
...
+	}
+
+	// Create the analytics tracker.
+	a := analytics.NewTracker(t, versions[version], osArch, runtime.GOOS, "SIGHUP", h)
+
+	defer a.Flush()
+
+	if _, err := cmd.NewRootCommand(versions, logW, a).ExecuteC(); err != nil {
+		panic(err)
```
omissis, line 72: in this case is not terrible but in general do not use panic unless you really need it.
go for a log + os.Exit(1) instead

main.go
```diff
@@ -41,11 +44,48 @@ func main() {
 		"osArch":    osArch,
 	}
 
-	rootCmd := cmd.NewRootCommand(versions, logFile)
+	logW, err := logFile()
+	if err != nil {
+		logrus.Fatal(err)
```
omissis, line 49:
```suggestion
		logrus.Error(err)
		os.Exit(1)
```

internal/analytics/tracker.go
```diff
@@ -0,0 +1,172 @@
...
+		a.events <- NewGuardEvent()
```
omissis, line 82: GuardEvent -> StopEvent

omissis (changes requested):

omissis (approved):

main.go
```diff
@@ -41,11 +44,49 @@ func main() {
...
+	a := analytics.NewTracker(t, versions[version], osArch, runtime.GOOS, "SIGHUP", h)
+
+	defer a.Flush()
+
+	if _, err := cmd.NewRootCommand(versions, logW, a).ExecuteC(); err != nil {
+		panic(err)
+	}
+}
+
+func logFile() (*os.File, error) {
```
Al-Pragliola, line 77: logFile function is not needed anymore in this file

main.go
```diff
@@ -41,11 +44,49 @@ func main() {
 		"osArch":    osArch,
 	}
 
-	rootCmd := cmd.NewRootCommand(versions, logFile)
+	logW, err := logFile()
```
Al-Pragliola, line 47: logW have to be removed and instead of that added a new variable:
```
var logFile *os.File
defer logFile.Close()
```

main.go
```diff
@@ -41,11 +44,49 @@ func main() {
...
+	if t == "" {
+		panic("FURYCTL_MIXPANEL_TOKEN environment variable not set")
 	}
+
+	// Create the analytics tracker.
+	a := analytics.NewTracker(t, versions[version], osArch, runtime.GOOS, "SIGHUP", h)
+
+	defer a.Flush()
+
+	if _, err := cmd.NewRootCommand(versions, logW, a).ExecuteC(); err != nil {
```
Al-Pragliola, line 72: pass logFile instead of logW as an argument to cmd.NewRootCommand

Al-Pragliola (changes requested):

Al-Pragliola (approved):

PR #110, opened 2022-11-02
fix(vendor): downgrade repository fallaback message to Info

ralgozino:
After seeing the reaction of a user that is not familiar with Fury on the WARN messages thrown by the fallback logic in the vendor command, I believe is better to downgrade them to INFO to avoid alarming the user unnecessary, at least until we decide to move on with the renaming.


Al-Pragliola (approved):
LGTM

omissis (approved):
LGTM

PR #113, opened 2022-11-03
fix: check if the current folder has the right write permissions

luigibarbato:
Running furyctl vendor -H inside a folder that the current user can't write to, shows absolutely no error and normal output but the vendor folder is nowhere to be found. So from now, furyctl check if the current folder has the right write permissions before to start the download process.

### DIFF
--- a/cmd/vendor.go
+++ b/cmd/vendor.go
@@ -5,9 +5,13 @@
 package cmd
 
 import (
+	"errors"
+	"os"
+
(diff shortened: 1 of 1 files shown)


luigibarbato:
the issue not present in furyctl 0.8.0. weh furyctl can't write to the folder it prints the proper error:

`ERRO[0028] mkdir vendor: permission denied`

PR #118, opened 2022-11-29
Feature/download bin default path

Al-Pragliola:


cmd/delete/cluster.go
```diff
@@ -46,6 +47,11 @@ func NewClusterCmd() *cobra.Command {
 				return ErrFuryctlFlagNotSet
 			}
 
+			distroLocation, ok := cobrax.Flag[string](cmd, "distro-location").(string)
+			if !ok {
+				return ErrDistroFlagNotSet
```
omissis, line 52: Wrap this error and add which flag is not set, same for all the flags below.
Perhaps there's a way to mark them as mandatory down below so there's no need to do this here?
It's also a bit weird that all these flags are mandatory, eg: dry-run and force should be defaulted and not required

cmd/delete/cluster.go
```diff
@@ -61,22 +67,40 @@ func NewClusterCmd() *cobra.Command {
...
 			homeDir, err := os.UserHomeDir()
 			if err != nil {
 				return fmt.Errorf("error while getting user home directory: %w", err)
 			}
 
-			minimalConf, err := yamlx.FromFileV3[config.Furyctl](furyctlPath)
+			// Init first half of collaborators.
+			client := netx.NewGoGetterClient()
+			executor := execx.NewStdExecutor()
+			distrodl := distribution.NewDownloader(client, debug)
```
omissis, line 79: these should be injected, and an interface should be used as type hint in the constructor.
Seeing we did this in many places, we could probably open a dedicated card to refactor it if we deem it necessary.

cmd/delete/cluster.go
```diff
@@ -61,22 +67,40 @@ func NewClusterCmd() *cobra.Command {
...
+
+			execx.Debug = debug
+
+			// Download the distribution.
+			logrus.Info("Downloading distribution...")
+			res, err := distrodl.Download(version, distroLocation, furyctlPath)
```
omissis, line 85: Is this "smart" and avoid downloading the distro if it's already present on the filesystem?

cmd/validate/dependencies.go
```diff
@@ -32,10 +32,6 @@ func NewDependenciesCmd(furyctlBinVersion string) *cobra.Command {
 			if !ok {
```
omissis, line 32: debug is mandatory?

internal/cluster/phase.go
```diff
@@ -41,17 +42,17 @@ type OperationPhaseOption struct {
 	Value any
 }
 
-func NewOperationPhase(folder string) (*OperationPhase, error) {
+func NewOperationPhase(folder string, kfdTools config.KFDTools) (*OperationPhase, error) {
 	basePath := folder
 
-	vendorPath, err := filepath.Abs(path.Join(basePath, "../", "vendor"))
+	vendorPath, err := filepath.Abs(path.Join(basePath, "../../", "bin"))
```
omissis, line 48:
```suggestion
	binPath, err := filepath.Abs(path.Join(basePath, "../../", "bin"))
```

internal/cluster/phase.go
```diff
@@ -41,17 +42,17 @@ type OperationPhaseOption struct {
...
-		return &OperationPhase{}, fmt.Errorf("error getting absolute path for vendor folder: %w", err)
+		return &OperationPhase{}, fmt.Errorf("error getting absolute path for tools bin folder: %w", err)
 	}
 
-	kustomizePath := path.Join(vendorPath, "bin", "kustomize")
-	terraformPath := path.Join(vendorPath, "bin", "terraform")
-	kubectlPath := path.Join(vendorPath, "bin", "kubectl")
+	kustomizePath := path.Join(vendorPath, "kustomize", kfdTools.Kustomize, "kustomize")
+	terraformPath := path.Join(vendorPath, "terraform", kfdTools.Terraform, "terraform")
+	kubectlPath := path.Join(vendorPath, "kubectl", kfdTools.Kubectl, "kubectl")
```
omissis, line 55:
```suggestion
	kustomizePath := path.Join(binPath, "kustomize", kfdTools.Kustomize, "kustomize")
	terraformPath := path.Join(binPath, "terraform", kfdTools.Terraform, "terraform")
	kubectlPath := path.Join(binPath, "kubectl", kfdTools.Kubectl, "kubectl")
```

omissis: I can't tell if those paths are correct but they look a bit odd, can you confirm they are the intended ones (repetition in the path seems weird, but maybe it's ok)?

internal/x/os/path.go
```diff
@@ -10,7 +10,18 @@ import (
 	"os"
 )
 
-func CleanupTempDir(dir string) error {
+type PathManager interface {
```
omissis, line 13: I think "Managers" are a strong design smell, as they are buckets of utilities functions with unclear boundaries and responsibilities (ie: when we don't know how to call something, it becomes a manager). In this case I suspect the implementation arose from the need of a test, but I would probably go the other way around and create a dedicated set of files and folders for that single test instead of spreading this implementation around the codebase so to support that one use case. That doesn't mean a Filesystem abstraction isn't useful per se, but here it doesn't seem the cost is justified, and the abstraction should probably be refined a bit more, as there are a lot of places where we still use `os` functions straight away.

omissis (changes requested):

internal/cluster/phase.go
Al-Pragliola: citing the issue: Default value proposal: $HOME/.furyctl/bin/dependency/version/<binary>

so the expected dir for a binary is: ~/.furyctl/bin/kubectl/1.23.7/ plus the binary => ~/.furyctl/bin/kubectl/1.23.7/kubectl

cmd/delete/cluster.go
Al-Pragliola: nope every time we download the distro to a temp folder

cmd/delete/cluster.go
Al-Pragliola: yep seems like a card is needed to plan this refactor

internal/x/os/path.go
Al-Pragliola: reverted and added a dedicated set of files

cmd/delete/cluster.go
omissis: @ralgozino thoughts on this?

cmd/validate/dependencies.go
omissis: @Al-Pragliola

cmd/delete/cluster.go
```diff
@@ -15,68 +15,85 @@ import (
...
 			if err != nil {
-				return fmt.Errorf("%w: %s", ErrYamlUnmarshalFile, err)
+				return fmt.Errorf("error while downloading distribution: %w", err)
 			}
 
-			basePath := filepath.Join(homeDir, ".furyctl", minimalConf.Metadata.Name)
+			basePath := filepath.Join(homeDir, ".furyctl", res.MinimalConf.Metadata.Name)
 
-			clusterDeleter, err := cluster.NewDeleter(minimalConf, phase, basePath)
+			binPath := filepath.Join(homeDir, ".furyctl", "bin")
```
omissis, line 85: shouldn't this be configurable via a flag? the validate dependency command handles that

internal/cluster/phase.go
```diff
@@ -41,17 +42,17 @@ type OperationPhaseOption struct {
 	Value any
 }
 
-func NewOperationPhase(folder string) (*OperationPhase, error) {
+func NewOperationPhase(folder string, kfdTools config.KFDTools) (*OperationPhase, error) {
 	basePath := folder
 
-	vendorPath, err := filepath.Abs(path.Join(basePath, "../", "vendor"))
+	binPath, err := filepath.Abs(path.Join(basePath, "../../", "bin"))
```
omissis, line 48: I am not sure this works every time, considering we are passing bin path as a flag in some commands

internal/dependencies/tools/tool.go
```diff
@@ -129,6 +131,17 @@ func (vc *checker) version(want string) error {
 		return errVersionEmpty
 	}
 
+	cmdDir := filepath.Dir(vc.runner.CmdPath())
+
+	_, err := os.Stat(cmdDir)
+	if err != nil {
```
omissis, line 137: nitpick
```suggestion
	if _, err := os.Stat(cmdDir);  err != nil {
```

omissis (changes requested):

cmd/validate/dependencies.go
Al-Pragliola: seems like it's not. It was a parameter in distro Downloader, but there was not used

cmd/delete/cluster.go
Al-Pragliola: true, as is bin-path is only used in validate dependencies. Now we can add bin-path to every command where binaries are needed, or remove bin-path from validation cmd

Al-Pragliola: after talking with @ralgozino we will add bin-path flag to every command where binaries are needed

cmd/delete/cluster.go
ralgozino: IMO as a first approach it is OK to download everytime and start from a clean state. The distro download is not a resource nor time intensive operation.

Once we have a clear overall picture we can start to optimize this kind of things.

omissis: ok

omissis (approved):

PR #119, opened 2022-12-01
feat: added --log flag

Al-Pragliola:


omissis (approved):
LGTM, I would consider one more refactoring step to bring most of the code here under `internal/x/log`

ralgozino (approved):
LGTM

PR #126, opened 2022-12-14
Feature/skip phase flag

Al-Pragliola:


cmd/create/cluster.go
```diff
@@ -151,6 +156,10 @@ func NewClusterCmd(version string, tracker *analytics.Tracker) *cobra.Command {
 				return fmt.Errorf("error while validating dependencies: %w", err)
 			}
 
+			if phase == "" || skipPhase == "distribution" {
+				vpnAutoConnect = true
+			}
```
omissis, line 161: Are you sure we want this behavior here? wouldn't it be less surprising if we left the choice to be controlled by the `vpn-auto-connect` flag alone?

omissis (commented):
Besides the couple of questions I left, I have one more about the tests data: it is extensive enough that we could maybe think of spending a minute going over alternatives, such as using the distribution code directly in tests if possible, as maintaining all those files will definitely have a cost.

internal/apis/kfd/v1alpha2/eks/creator.go
```diff
@@ -132,18 +132,23 @@ func (v *ClusterCreator) Create(dryRun bool) error {
 		return nil
 
 	case cluster.OperationPhaseAll:
-		if v.furyctlConf.Spec.Infrastructure != nil {
+		if v.furyctlConf.Spec.Infrastructure != nil &&
+			(skipPhase == "" || skipPhase == cluster.OperationPhaseDistribution) {
```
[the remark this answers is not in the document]
Al-Pragliola: skipPhase == "" => doing all the phases so infra included
skipPhase == "kubernetes" => skipping also infra
skipPhase == "distribution" => doing infra and then kube but not distribution

with skipPhase != cluster.OperationPhaseInfrastructure we get the skipPhase == "kubernetes" wrong

cmd/create/cluster.go
Al-Pragliola: depends if we want to force the user to set vpnAutoConnect when doing the phase == "" (all phases) and there's a VPN in the furyctl infra vpc section, if the user forgets this flag then the kubernetes phase fails badly with the only option to delete sgs,eks,... via other tools

omissis (approved):

PR #127, opened 2022-12-16
feat: add awscli's tool and runner

luigibarbato:
Waiting for https://github.com/sighupio/fury-distribution/pull/33


.gitignore
```diff
@@ -14,3 +14,6 @@ test/e2e/e2e.test
 
 .vscode
 *.log
+go.work
+go.work.sum
+.env
```
omissis, line 19:
```suggestion
.env

```

internal/dependencies/tools/awscli.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func (*Awscli) SupportsDownload() bool {
+	return false
+}
+
+func (f *Awscli) SrcPath() string {
+	if f.os == "darwin" {
+		return fmt.Sprintf(
+			"https://awscli.amazonaws.com/AWSCLIV2-%s.pkg",
```
omissis, line 51: This function should not be implemented as we don't support aws cli downloads.

internal/dependencies/tools/awscli.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return ""
+}
+
+func (f *Awscli) Rename(basePath string) error {
+	oldName := fmt.Sprintf("awscli-%s-%s", f.os, f.arch)
+	newName := "aws-cli"
+
+	err := os.Rename(filepath.Join(basePath, oldName), filepath.Join(basePath, newName))
+	if err != nil {
```
omissis, line 73: nitpick
```suggestion
	if err := os.Rename(filepath.Join(basePath, oldName), filepath.Join(basePath, newName)); err != nil {
```

internal/dependencies/tools/test_data/aws-cli/2.8.12/aws
```diff
@@ -0,0 +1,27 @@
+#!/opt/homebrew/Cellar/awscli/2.8.12/libexec/bin/python3.10
```
omissis, line 1: this is macos only, it should be a bit more generic I think

omissis: Also: what is this used for? perhaps a simpler shell script that outputs a hard-coded version as we did in other tests is good enough?

internal/dependencies/tools/validator.go
```diff
@@ -37,17 +37,19 @@ func (tv *Validator) Validate(kfdManifest config.KFD) []error {
 
 	tls := reflect.ValueOf(kfdManifest.Tools)
 	for i := 0; i < tls.NumField(); i++ {
-		name := strings.ToLower(tls.Type().Field(i).Name)
+		for j := 0; j < tls.Field(i).NumField(); j++ {
+			name := strings.ToLower(tls.Field(i).Type().Field(j).Name)
```
omissis, line 41: this can be moved down below before `tool := tv.toolFactory.Create(name, version.String())` and save some computation in case the two ifs below go wrong

internal/dependencies/tools/awscli.go
```diff
@@ -0,0 +1,106 @@
...
+			f.os,
+			f.arch,
+		)
+	}
+
+	return ""
+}
+
+func (f *Awscli) Rename(basePath string) error {
+	oldName := fmt.Sprintf("awscli-%s-%s", f.os, f.arch)
```
omissis, line 69: This function should not be implemented as we don't support aws cli downloads.

internal/dependencies/tools/awscli.go
```diff
@@ -0,0 +1,106 @@
...
+
+func (f *Awscli) CheckBinVersion() error {
+	if err := f.checker.version(f.version); err != nil {
+		return fmt.Errorf("aws-cli: %w", err)
+	}
+
+	return nil
+}
+
+func getSrcPath(os, arch, version string) string {
```
omissis, line 88: remove this function as it is never used.

internal/dependencies/download.go
```diff
@@ -133,35 +133,40 @@ func (dd *Downloader) DownloadTools(kfdTools config.KFDTools) ([]string, error)
...
+			if err != nil {
+				return unsupportedTools, fmt.Errorf("%w '%s': %v", distribution.ErrChangingFilePermissions, tool.SrcPath(), err)
+			}
 		}
 	}
 
 	return unsupportedTools, nil
 }
+
+func getTools() {}
```
omissis, line 172: remove as it was never used.

internal/dependencies/tools/awscli_test.go
```diff
@@ -0,0 +1,138 @@
...
+func Test_Awscli_SupportsDownload(t *testing.T) {
+	a := tools.NewAwscli(newAwscliRunner(), "0.3.0")
+
+	if a.SupportsDownload() != false {
+		t.Errorf("Awscli download must not be supported")
+	}
+}
+
+func Test_Awscli_SrcPath(t *testing.T) {
+	var wantSrcPath string
```
omissis, line 31: this should only test an empty string is returned

internal/dependencies/tools/awscli_test.go
```diff
@@ -0,0 +1,138 @@
...
+			fa := tools.NewAwscli(newAwscliRunner(), tC.version)
+			if fa.SrcPath() != wantSrcPath {
+				t.Errorf("Wrong Awscli src path: want = %s, got = %s", wantSrcPath, fa.SrcPath())
+			}
+		})
+	}
+}
+
+func Test_Awscli_Rename(t *testing.T) {
+	tmpDir, err := os.MkdirTemp("", "furyctl-test-")
```
omissis, line 65: this should only test an the method returns nil and does nothing

omissis (changes requested):

internal/apis/kfd/v1alpha2/eks/create/infrastructure.go
```diff
@@ -140,7 +144,7 @@ func (i *Infrastructure) Exec(opts []cluster.OperationPhaseOption) error {
 		}
 
 		for _, opt := range opts {
-			if strings.ToLower(opt.Name) == cluster.OperationPhaseOptionVPNAutoConnect {
+			if strings.ToLower(opt.Name) == cluster.OperationPhaseOptionVPNAutoConnect && opt.Value == true {
```
omissis, line 147: please revert this change as it's fixed in here (PR #126): https://github.com/sighupio/furyctl/pull/126/files#diff-8a6e19e5142516cbc6e52987fe939bf5beaac01b413e7141eca23a9a5c2c83dfR144

omissis (changes requested):

internal/dependencies/tools/test_data/aws-cli/2.8.12/aws
luigibarbato: this is the aws-cli's binary. We use the binary for the unit tests as other tools

omissis (approved):

PR #128, opened 2022-12-20
Handling private or non exist modules

luigibarbato:
Furyctl should properly handle private and non-existing repositories as done in the furyctl legacy version


Al-Pragliola (approved):
LGTM

PR #129, opened 2022-12-20
feat!(eks): add support for EKS Installer v1.10.0

ralgozino:
- add missing mappings for EKS  ClusterLogRetentionDays parameter
- expose new parameter for EKS NodePoolsLaunchKind
- Enforce that NodePoolsLaunchKind is set and not defaulted to avoid node pools rollouts
- improve EKS cluster console messages
- update test to consider the new parameters
- update docs to prepare for releasing v0.9.0
- fix Kubeconfig generation. v1alpha1 is deprecated.

BREAKING CHANGE: there are two 2 parameters in the configuration file, ClusterLogRotationDays can be left as default in most cases, but users need to set explicitly the new NodePoolsLaunchKind parameter, because migrating to the new kind requires a rollout of all the existing node pools.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ You can find `furyctl` binaries on the [Releases page](https://github.com/sighup
 To download the latest release, run:
 
 ```bash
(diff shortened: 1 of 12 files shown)


lnovara (approved):
LGTM :shipit:

omissis (approved):

PR #131, opened 2023-01-02
Feature/first wave adjustments

Al-Pragliola:
Changelist:

- add debug message when waiting for cluster resources removal
- append '...' to Deleting x phase messages
- when deleting a cluster or deleting just phase infrastructure print a WARN telling the user to kill the openvpn process
- add a --kubeconfig flag (mandatory if running distribution phase alone and $KUBECONFIG is not set) to `create cluster`
- add a --kubeconfig flag (mandatory if running distribution phase alone or all phases and $KUBECONFIG is not set) to `delete cluster`
- bumped up golang version to 1.19.4, golangci-lint to 1.50.1
- refactored functions with too many parameters

### DIFF
--- a/cmd/create/cluster.go
+++ b/cmd/create/cluster.go
@@ -28,7 +28,22 @@ import (
 var (
 	ErrParsingFlag                = errors.New("error while parsing flag")
 	ErrDownloadDependenciesFailed = errors.New("download dependencies failed")
+	ErrKubeconfigReq              = errors.New("$KUBECONFIG is not set, so --kubeconfig is required when doing distribution phase alone")
 )
+
+type ClusterCmdFlags struct {
+	Debug              bool
+	FuryctlPath        string
+	DistroLocation     string
+	Phase              string
+	SkipPhase          string
+	BinPath            string
+	VpnAutoConnect     bool
+	DryRun             bool
+	SkipDepsDownload   bool
+	SkipDepsValidation bool
+	Kubeconfig         string
+}
 
 func NewClusterCmd(version string, tracker *analytics.Tracker) *cobra.Command {
 	var cmdEvent analytics.Event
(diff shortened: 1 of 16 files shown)


cmd/delete/cluster.go
```diff
@@ -68,12 +71,30 @@ func NewClusterCmd(version string, tracker *analytics.Tracker) *cobra.Command {
...
+			// Check if kubeconfig is needed.
+			if phase == "distribution" || phase == "" {
+				if kubeconfig == "" {
+					kubeconfigFromEnv := os.Getenv("KUBECONFIG")
+
+					if kubeconfigFromEnv == "" {
+						return ErrKubeconfigReq
+					}
+
+					logrus.Warnf("Missing --kubeconfig flag, fallback to KUBECONFIG from environment: %s", kubeconfigFromEnv)
```
ralgozino, line 94: Same as in create

cmd/delete/cluster.go
```diff
@@ -198,6 +226,13 @@ func NewClusterCmd(version string, tracker *analytics.Tracker) *cobra.Command {
 		"Force deletion of the cluster",
 	)
 
+	cmd.Flags().String(
+		"kubeconfig",
+		"",
+		"Path to the kubeconfig file, mandatory if you want to run the distribution phase alone or "+
+			"if you want to delete a cluster and $KUBECONFIG is not set",
```
ralgozino, line 233:
```suggestion
			"if you want to delete a cluster and the KUBECONFIG environment variable is not set",
```

internal/apis/kfd/v1alpha2/eks/delete/infrastructure.go
```diff
@@ -49,19 +48,22 @@ func NewInfrastructure(dryRun bool, workDir, binPath string, kfdManifest config.
...
 	}
 
 	err = i.tfRunner.Destroy()
 	if err != nil {
-		return fmt.Errorf("error running terraform destroy: %w", err)
+		return fmt.Errorf("error while deleting infrastructure: %w", err)
 	}
 
+	logrus.Warnf("Please, remember to kill the OpenVPN process if" +
+		" you have chosen to create it in the infrastructure phase")
```
ralgozino, line 66: can we add a command here so the user can copy & paste it?

ralgozino (changes requested):
I can't say much on the Go code, but I left some UX suggestions

cmd/create/cluster.go
```diff
@@ -97,8 +70,21 @@ func NewClusterCmd(version string, tracker *analytics.Tracker) *cobra.Command {
...
+			// Check if kubeconfig is needed.
+			if flags.Phase == "distribution" || flags.SkipPhase == "kubernetes" {
+				if flags.Kubeconfig == "" {
+					kubeconfigFromEnv := os.Getenv("KUBECONFIG")
+
+					if kubeconfigFromEnv == "" {
+						return ErrKubeconfigReq
+					}
+
+					logrus.Warnf("Missing --kubeconfig flag, fallback to KUBECONFIG from environment: %s", kubeconfigFromEnv)
```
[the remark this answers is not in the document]
Al-Pragliola: The idea of the Warn was to focus more on which cluster the user was going to apply/delete resources, on the distribution phase alone, but it's not strong reasoning.

internal/apis/kfd/v1alpha2/eks/delete/infrastructure.go
Al-Pragliola: Yep, do you think it's better to do that when we work on https://github.com/sighupio/furyctl/issues/140 ?
to be sure about needing sudo or not when killing the process

ralgozino: makes sense 👍

cmd/create/cluster.go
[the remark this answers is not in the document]
ralgozino: OK, let's leave it as warn and see what user feedback tell us. I'd suggest changing `fallback` to `fallingback` in the message though

ralgozino (approved):
LGTM

PR #147, opened 2023-01-03
feat: removed furyctl binary version check

Al-Pragliola:
Changelist:

- removed function `checkFuryctlBinVersion` from the dependencies validation step and refactored accordingly


ralgozino (approved):
LGTM

PR #149, opened 2023-01-04
feat: handle unsupported phases

Al-Pragliola:
Changelist:

- added an error when trying to use an unsupported phase flag in create/delete cluster command:
`ERRO error while parsing flag: phase: options are: infrastructure, kubernetes, distribution`
- added an error when trying to use an unsupported skip-phase flag in create cluster command:
`ERRO error while parsing flag: skip-phase: unsupported phase, options are: infrastructure, kubernetes, distribution`
- added an error when trying to use phase and skip-phase flag together:
`ERRO error while parsing flag: skip-phase: cannot use together with phase flag`
- refactored "infrastructure", "kubernetes", "distribution" to use enums instead

### DIFF
--- a/cmd/create/cluster.go
+++ b/cmd/create/cluster.go
@@ -71,7 +71,7 @@ func NewClusterCmd(tracker *analytics.Tracker) *cobra.Command {
 			}
 
 			// Check if kubeconfig is needed.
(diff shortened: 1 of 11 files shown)


ralgozino (approved):
lgtm

PR #156, opened 2023-01-11
chore: check github token whitespaces, improve logrus format

luigibarbato:
After https://github.com/sighupio/fury-getting-started/issues/25 (thanks @raclepoulpe) we discovered a bug in the go-getter's library that occurs when the passed URL has whitespaces in it. In this case, the GITHUB_TOKEN erroneously presented a whitespace injected into the OAuth URI to be given to the go-getter's client. This PR is intended to partially mitigate this problem by warning the user with a warning message.

### DIFF
--- a/cmd/vendor.go
+++ b/cmd/vendor.go
@@ -5,13 +5,16 @@
 package cmd
 
 import (
+	"os"
+	"strings"
+
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
 
 func init() {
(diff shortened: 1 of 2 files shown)


cmd/vendor.go
```diff
@@ -55,6 +58,10 @@ var vendorCmd = &cobra.Command{
 			return err
 		}
 
+		if token := os.Getenv("GITHUB_TOKEN"); token != "" && strings.Contains(token, " ") {
```
omissis, line 61: `token != "" &&` seems redundant, as an empty string cannot contain a space

cmd/vendor.go
```diff
@@ -55,6 +58,10 @@ var vendorCmd = &cobra.Command{
...
+			logrus.Warn("GITHUB_TOKEN seems to be malformed. Vendoring modules may fail. Please double check its content.\n")
```
omissis, line 62:
```suggestion
			logrus.Warn("GITHUB_TOKEN contains a space character. As a result, vendoring modules may fail. If it's intended, you can ignore this warning.\n")
```

omissis (changes requested):

ralgozino (approved):
lgtm

PR #157, opened 2023-01-12
feat: added dry-run logic in phase distribution to create cluster cmd

Al-Pragliola:
Changelist:

- distribution(dry-run): output terraform plan and kustomize build

example:
demo


ralgozino (approved):
From the screen capture and the code diff, LGTM

PR #158, opened 2023-01-12
chore: remove logrus timestamp format

luigibarbato:


main.go
```diff
@@ -49,16 +49,26 @@ func exec() int {
...
+			DisableTimestamp: true,
+		},
+	}
+
+	log.SetLevel(logrus.InfoLevel)
+
 	h, err := os.Hostname()
 	if err != nil {
-		logrus.Debug(err)
+		log.Debug(err)
```
Al-Pragliola, line 64: after setting `log.SetLevel(logrus.InfoLevel)` we'll never see this debug message

main.go
```diff
@@ -49,16 +49,26 @@ func exec() int {
...
 
 		h = "unknown"
 	}
 
 	t := os.Getenv("FURYCTL_MIXPANEL_TOKEN")
 	if t == "" {
-		logrus.Debug("FURYCTL_MIXPANEL_TOKEN is not set")
+		log.Debug("FURYCTL_MIXPANEL_TOKEN is not set")
```
Al-Pragliola, line 71: after setting `log.SetLevel(logrus.InfoLevel)` we'll never see this debug message

Al-Pragliola (commented):
@ralgozino @luigibarbato

main.go
luigibarbato: Should we rethink how we handle logrus logs in the cmd?

main.go
luigibarbato: Should we rethink how we handle logrus logs in the cmd?

Al-Pragliola: mm maybe the whole logic is wrong, why are we doing that in the main.go?
```
	// Create the analytics tracker.
	a := analytics.NewTracker(t, versions[version], osArch, runtime.GOOS, "SIGHUP", h)
```

when we have in root.go the flag disable-analytics, shouldn't we wait for the flag to be parsed before creating a tracker?

luigibarbato:
Force the Debug level as a workaround for now but we'll have to rethink how we handle logging in to the codebase.

Al-Pragliola (approved):
LGTM

PR #171, opened 2023-01-19
Feat: clean up vendor folder before dl and remove .git subfolder after dl

### LINKED ISSUE #167: [Furyctl next] when downloading on vendor, remove all .git folders
When downloading, we need to delete the target folder of the module, re-download and delete all .git folders.

Al-Pragliola:
Changelist:

- added folder removal before downloading a module/installer
- added .git subfolder removal after downloading a module/installer
 
closes #167

### DIFF
--- a/internal/dependencies/download.go
+++ b/internal/dependencies/download.go
@@ -14,6 +14,8 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
 	"github.com/sighupio/fury-distribution/pkg/config"
 	"github.com/sighupio/furyctl/internal/dependencies/tools"
 	"github.com/sighupio/furyctl/internal/distribution"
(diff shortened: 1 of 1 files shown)


ralgozino (changes requested):
Can we add some DEBUG message sayin what we are deleting?

From my tests, seems that we are deleting only the folders for the things that we download instead of the whole vendor folder.

For example, when we download a module x, we delete the folder `<cluster_name>/vendor/modules/x` instead of cleaning up the whole vendor folder before starting the download.

This could leave garbage behind, for example, let's say that we don't include a module anymore in a new version. When we download the new version all the files for the deprecated module will remain.

The only thing that we should consider when deleting the whole folder is what happens is the downloads fail for some reason, then we won't have the new files nor the old ones anymore.

Thoughts @Al-Pragliola @nutellinoit ?

Al-Pragliola:
> Can we add some DEBUG message sayin what we are deleting?
> 
> From my tests, seems that we are deleting only the folders for the things that we download instead of the whole vendor folder.
> 
> For example, when we download a module x, we delete the folder `<cluster_name>/vendor/modules/x` instead of cleaning up the whole vendor folder before starting the download.
> 
> This could leave garbage behind, for example, let's say that we don't include a module anymore in a new version. When we download the new version all the files for the deprecated module will remain.
> 
> The only thing that we should consider when deleting the whole folder is what happens is the downloads fail for some reason, then we won't have the new files nor the old ones anymore.
> 
> Thoughts @Al-Pragliola @nutellinoit ?

We decided to remove the vendor folder before installing modules/installers, in order to have a clean state and be safe on possible problems with old files

ralgozino (approved):
LGTM

PR #172, opened 2023-01-19
Feat: check cluster connectivity in distribution phase

### LINKED ISSUE #168: [Furyctl next] In distribution phase, error if kubectl version does not answer a server version

Al-Pragliola:
Changelist:

- added connectivity check into cluster create phase distribution by running kubectl version without flag --client 
- added connectivity check into cluster delete phase distribution by running kubectl version without flag --client 

closes #168

### DIFF
--- a/internal/tool/kubectl/runner.go
+++ b/internal/tool/kubectl/runner.go
@@ -152,8 +152,18 @@ func (r *Runner) Delete(manifestPath string, params ...string) error {
 }
 
 func (r *Runner) Version() (string, error) {
+	args := []string{"version"}
+
+	if r.paths.Kubeconfig != "" {
+		args = append(args, "--kubeconfig", r.paths.Kubeconfig)
+	}
+
+	if !r.serverSide {
+		args = append(args, "--client")
+	}
+
 	out, err := execx.CombinedOutput(execx.NewCmd(r.paths.Kubectl, execx.CmdOptions{
-		Args:     []string{"version", "--client"},
+		Args:     args,
 		Executor: r.executor,
 		WorkDir:  r.paths.WorkDir,
 	}))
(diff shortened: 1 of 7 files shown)


internal/tool/kubectl/runner.go
```diff
@@ -152,8 +152,18 @@ func (r *Runner) Delete(manifestPath string, params ...string) error {
...
 
 func (r *Runner) Version() (string, error) {
+	args := []string{"version"}
+
+	if r.paths.Kubeconfig != "" {
+		args = append(args, "--kubeconfig", r.paths.Kubeconfig)
+	}
+
+	if !r.serverSide {
+		args = append(args, "--client")
```
ralgozino, line 162: Correct me if I'm wrong, but this logic is not correct. `serverSide` is used for example, when you do `kubectl apply --server-side`, but it is not correct to assume that if I want to do a server side apply I will also want to run `kubectl version` without the `--client` flag.

The two flags are not related.

I'd suggest separating the properties.

ralgozino (changes requested):

internal/tool/kubectl/runner.go
Al-Pragliola: _haste makes waste_

ralgozino (changes requested):
From my tests, the connectivity check is being run with the `--client` flag set. So, it doesn't fail when it can't connect to the server.

See internal/tool/runner.go#78

The `--client` is useful when checking kubectl's version but we should disable it when doing the connectivity check.

Also, I'd suggest using the `--output=json` flag in the `version` command, so we get a nice parsable output :)

The --ouput=json option has been present at least since 1.20, so it is safe to use it.

Al-Pragliola:
> From my tests, the connectivity check is being run with the `--client` flag set. So, it doesn't fail when it can't connect to the server.
> 
> See internal/tool/runner.go#78
> 
> The `--client` is useful when checking kubectl's version but we should disable it when doing the connectivity check.
> 
> Also, I'd suggest using the `--output=json` flag in the `version` command, so we get a nice parsable output :)
> 
> The --ouput=json option has been present at least since 1.20, so it is safe to use it.

added new issue about changing output in kubectl version: https://github.com/sighupio/furyctl/issues/178

ralgozino (approved):
LGTM

PR #173, opened 2023-01-19
Feat: UX improvement about unsupported KFD version

### LINKED ISSUE #165: [Furyctl Next] UX: unsupported KFD version
## What we did

- created a config file with `furyctl create config --distro-location=<pointing to feature/furyctl-next branch> --version 1.24.0`
- filled the generated config file
- run `furyctl validate config`

## What happened

We got an error saying that the kfd.yaml file did not exist.

## What should happen

Furyctl NG instead of failing with an error that kfd.yaml does not exists for commands that download the distribution, it should check if the needed file (kfd.yaml) exist and if it doesn't inform the user:
- that the specified version (print the version) is not compatible with furyctl ng.
- that they can use the `--distro-location` flag for development purposes.

Al-Pragliola:
Changelist:

- replaced missing `kfd.yaml` message with info about unsupported KFD version when running with or without --distro-location flag

closes #165

### DIFF
--- a/internal/distribution/download.go
+++ b/internal/distribution/download.go
@@ -19,7 +19,7 @@ import (
 	yamlx "github.com/sighupio/furyctl/internal/x/yaml"
 )
 
-const DefaultBaseURL = "https://git@github.com/sighupio/fury-distribution?ref=%s"
+const DefaultBaseURL = "git::git@github.com:sighupio/fury-distribution?ref=%s"
 
 var (
 	ErrChangingFilePermissions = errors.New("error changing file permissions")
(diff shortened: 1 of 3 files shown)


ralgozino (approved):
LGTM 🚀

ralgozino (changes requested):
sorry for the last-minute change. But I've just noticed that with this change, if I put a non-existing version in the `--version` flag, the error I get is misleading:

```console
go run main.go create config --version v1.25.0
DEBU FURYCTL_MIXPANEL_TOKEN is not set
ERRO failed to download distribution: unsupported KFD version: v1.25.0 is not supported by furyctl-ng, try another version or use flag --distro-location to specify a custom location
exit status 1
```

Al-Pragliola:
> 

now handling errors when kfd version does not exist:
Screenshot 2023-01-20 at 11 09 57

ralgozino (approved):
perfect <3

PR #176, opened 2023-01-20
Feat: add depth=1 to go-getter urls

### LINKED ISSUE #169: [Furyctl next] verify if it's possible to have go getter clone with shallow depth 1

Al-Pragliola:
Changelist:

- added depth=1 to every go-getter url

⚠️ needs #171 to be merged

NOTE: this is possibly breaking, now it's not possible to specify a commit hash as a ref
Screenshot 2023-01-20 at 12 02 31

closes #169


ralgozino (approved):
LGTM.

I think that the win in speed and disk usage is better than the losing the possibility of specifying a commit id, tag and branch should be enough for everybody™ (famous last words)

PR #194, opened 2023-02-01
Feat: only use sudo if necessary

### LINKED ISSUE #140: [Furyctl Next] only use sudo if necessary
Furyctl Next uses `sudo` to setup the openvpn connection, but without doing any prior checks.

It shall use `sudo` only when needed, for example, if the user is not `root`.

Al-Pragliola:
Changelist:

- added isRoot function to check if the current user is root
- modified all log messages/commands that were dependent from sudo
- added info at the end of cluster creation/infra creation about killing the VPN connection after finishing doing operations on the cluster

closes #140

### DIFF
--- a/cmd/delete/cluster.go
+++ b/cmd/delete/cluster.go
@@ -115,6 +115,7 @@ func NewClusterCmd(tracker *analytics.Tracker) *cobra.Command {
 			clusterDeleter, err := cluster.NewDeleter(
 				res.MinimalConf,
 				res.DistroManifest,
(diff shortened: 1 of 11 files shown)


ralgozino (changes requested):

ralgozino (commented):
The PR is looking good, I think we should also print a message at the end of the creation process reminding the user to kill the VPN after they have finished

(Maybe we already do that and I don't remember)

ralgozino (approved):
lgtm

PR #202, opened 2023-02-07
Create arm64 release packages

g-iannelli:
Add `arm64` architecture when creating release packages


Al-Pragliola (approved):
closes #203 

LGTM

PR #218, opened 2023-02-07
Feat: point source to local folder instead of remote

### LINKED ISSUE #214: infrastructure and kubernetes' terraform templates should use the local vendored installer folder
We should change the source in the terraform template from Github to local vendor folder

Al-Pragliola:
closes #214


ralgozino (approved):
lgtm

PR #221, opened 2023-02-09
feat(bootstrap,cluster): add --upgrade-deps flag

ralgozino:
## Summary

Add `--upgrade-deps` flag to bootstrap apply and cluster apply commands, allowing to upgrade terraform providers version.

## Usage
Assume you have an existing cluster, upgrading it to a newer version of the installer may require upgrading also the underlaying terraform providers version.

For example, in an EKS cluster you might need to upgrade the eks terraform provider version.

In plain terraform you'd do this by running `terraform init -upgrade`. This is not currently supported by furyctl.

This PR adds a `--ugprade-deps` flag to the `bootstrap apply` and `cluster apply` commands, so setting this flag makes furyctl pass the `-upgrade` flag to the `terraform init` command that gets executed when you do `{boostrap,cluster} apply` effectively upgrading the providers.

## Recommended use flow

0. Make a backup of the `bootstrap` and `cluster` folders.

1. update the bootstrap providers:

```bash
furyctl bootstrap apply --dry-run --upgrade-deps
```

2. check the output in `bootstrap/logs/terraform.log` and apply the changes if everything is OK:

```bash
furyctl bootstrap apply
```

3. update the cluster providers

```bash
furyctl cluster apply --dry-run --upgrade-deps
```
4. check the output in `cluster/logs/terraform.log` and apply the changes if everything looks OK:

```bash
furyctl cluster apply
```


nutellinoit (approved):
LGTM, will this be ported also on the new furyctl?

ralgozino:
> LGTM, will this be ported also on the new furyctl?

IMO right now there's nothing to port to ng. We should keep this in mind though when we design the `upgrade` feature.

Al-Pragliola (approved):
LGTM

PR #271, opened 2023-02-15
Fix: check that the infra section exists before infra phase

### LINKED ISSUE #200: Handle no infrastructure with --phase infrastructure in create/delete cmd
Furyctl should throw an error if create/delete cluster cmd is executed with --phase infrastructure but there is no infrastructure in the furyct.yaml file.

Al-Pragliola:
fixes #200

### DIFF
--- a/internal/apis/kfd/v1alpha2/eks/creator.go
+++ b/internal/apis/kfd/v1alpha2/eks/creator.go
@@ -24,7 +24,10 @@ import (
 	kubex "github.com/sighupio/furyctl/internal/x/kube"
 )
 
-var ErrUnsupportedPhase = errors.New("unsupported phase")
+var (
+	ErrUnsupportedPhase = errors.New("unsupported phase")
+	ErrInfraNotPresent  = errors.New("cluster configuration does not contain infrastructure section")
+)
 
 type ClusterCreator struct {
 	paths          cluster.CreatorPaths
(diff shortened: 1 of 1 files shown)


internal/apis/kfd/v1alpha2/eks/creator.go
```diff
@@ -24,7 +24,10 @@ import (
 	kubex "github.com/sighupio/furyctl/internal/x/kube"
 )
 
-var ErrUnsupportedPhase = errors.New("unsupported phase")
+var (
+	ErrUnsupportedPhase = errors.New("unsupported phase")
+	ErrInfraNotPresent  = errors.New("cluster configuration does not contain infrastructure section")
```
ralgozino, line 29:
```suggestion
	ErrInfraNotPresent  = errors.New("configuration file does not contain infrastructure section")
```

if we could add also the path to the file, that would be even better

ralgozino (changes requested):

internal/apis/kfd/v1alpha2/eks/creator.go
Al-Pragliola: now is something like:
```
ERRO error while creating cluster: the configuration file does not contain an infrastructure section: check at /path/to/furyctl.yaml
```

ralgozino (approved):
lgtm

PR #300, opened 2023-02-22
Chore: removed warning about cluster reachability when in dry-run

### LINKED ISSUE #237: Invalid error log on dry-run before creating the cluster
On first `furyctl create cluster --dry-run` why I got the following log message? 

```shell
WARN Cluster is unreachable, make sure it is reachable before running the command without --dry-run
```

Al-Pragliola:
Changelist:

- removed cluster reachability warning when running --dry-run without --phase
- phase distribution will not return an error when terraform plan fails in --dry-run
 
closes #237


ralgozino (approved):
SBAM

PR #307, opened 2023-02-24
feat: introduce extra validation beyond json schema

omissis:


internal/config/validate.go
```diff
@@ -106,7 +107,32 @@ func Validate(path, repoPath string) error {
...
+
+	err = validateSchemaExtraRules(path)
+	if err != nil {
+		return fmt.Errorf("error while validating against schema rules: %w", err)
+	}
+
+	return nil
+}
+
+func validateSchemaExtraRules(confPath string) error {
```
Al-Pragliola, line 121: I don't really like this function to be here because config/validate was not bound to a specific apiVersion but for now, this does the trick

omissis: @Al-Pragliola you are right, let's move it to a per-api implementation.

Al-Pragliola (approved):
LGTM, the only thing that comes to my mind is that schema validation doesn't fail on the first error
```
ERRO error while validating configuration file: error validating config: Key: 'Furyctl.Metadata.Name' Error:Field validation for 'Name' failed on the 'required' tag
Key: 'Furyctl.Spec.ToolsConfiguration.Terraform.State.S3.BucketName' Error:Field validation for 'BucketName' failed on the 'required' tag
Key: 'Furyctl.Spec.ToolsConfiguration.Terraform.State.S3.KeyPrefix' Error:Field validation for 'KeyPrefix' failed on the 'required' tag
Key: 'Furyctl.Spec.ToolsConfiguration.Terraform.State.S3.Region' Error:Field validation for 'Region' failed on the 'required' tag
```

while on the extra validation, we get one error at a time

omissis:
@Al-Pragliola schema validation does suffer from the same issue, I opened a new issue to track that improvement: #311

Issue #283, opened 2023-02-21
Wrong `nodeSelectors` for infra pods


g-iannelli:
The following resources have wrong nodeSelector:

|Namespace|Name|Kind|
|---|---|---|
|`kube-system`|`coredns`|`Deployment`|
|`calico-system`| `calico-kube-controllers`|`Deployment`|
|`monitoring`|`x509-certificate-exporter-data-plane`|`Daemonset`|
|`monitoring`|`x509-certificate-exporter`|`Deployment`|
|`logging`| `kubernetes-event-tailer`|`StatefulSet` (operated?)|
|`logging`|`ism-policy-cronjob`|`CronJob`|
|`logging`|`index-patterns-cronjob`|`CronJob`|
|`logging` |`logging-operator`|`Deployment`|
|`gatekeeper-system`| `gatekeeper-policy-manager`|`Deployment`|

This issue should replace the #264

omissis:
All the patches should be doable using kustomize except the `coredns` one, which needs to be applied directly using `kubectl` at the end of the`kubernetes` phase.
This requires to add the following section within the kubernetes section:
```
    common:
      nodeSelector:
        node.kubernetes.io/role: infra
```
The schema needs to be updated as well

omissis:
another potential approach for the short term would be to use kubectl as part of a null_resource

omissis:
It should be, as it's the only one that requires that kind of workaround

PR #327, opened 2023-03-21
feat: added coredns patch on phase kubernetes

Al-Pragliola:
Changelist:

- added furyctl config merge also on kubernetes phase
- added coredns patch on terraform phase kubernetes
- fixed on distribution ignore folders

partially fixes: #283

### DIFF
--- a/configs/provisioners/cluster/eks/coredns.tf.tpl
+++ b/configs/provisioners/cluster/eks/coredns.tf.tpl
@@ -1,0 +1,54 @@
+locals {
+  coredns_scheduling_patch = {
+    spec = {
+      template = {
+        spec = {
+          nodeSelector = {{ if hasKeyAny .distribution "nodeSelector" }}{
+            {{- range $key, $value := .distribution.nodeSelector }}
+            "{{ $key }}" = "{{ $value }}"
+            {{- end }}
+          } {{ else }} null {{ end }}
+          tolerations = [
+            {{- range $key, $value := .distribution.tolerations }}
+            {
+              key = "{{ $value.key }}"
+              value = "{{ $value.value }}"
+              effect = "{{ $value.effect }}"
+            },
+            {{- end }}
+          ]
+        }
+      }
+    }
+  }
+ coredns_scheduling_patch_as_json = jsonencode(local.coredns_scheduling_patch)
+}
+
+resource "local_file" "cluster_ca" {
+   depends_on = [
+     module.fury
+   ]
+
+  content = base64decode(data.aws_eks_cluster.fury.certificate_authority.0.data)
+  filename = "${path.module}/secrets/ca.crt"
+}
+
+resource "null_resource" "patch_coredns" {
+  depends_on = [
+    module.fury,
+    local_file.cluster_ca
+  ]
+
+  triggers = {
+    run_once = local.coredns_scheduling_patch_as_json
+  }
+
+  provisioner "local-exec" {
+    command = <<-EOT
+      ${var.kubectl_path} patch deployment/coredns -n kube-system -p '${local.coredns_scheduling_patch_as_json}' \
+      --server=${data.aws_eks_cluster.fury.endpoint} \
+      --token=${data.aws_eks_cluster_auth.fury.token} \
+      --certificate-authority=${path.module}/secrets/ca.crt \
+      EOT
+  }
+}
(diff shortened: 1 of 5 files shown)


g-iannelli:
The patch must be applied on `kubernetes` phase, after cluster creation

g-iannelli (changes requested):
https://github.com/sighupio/fury-distribution/pull/65 has been closed, so this PR shoul be reviewed to move the `kubectl_path` into `kubernetes` phase.

g-iannelli (changes requested):

configs/provisioners/cluster/eks/coredns.tf.tpl
```diff
@@ -0,0 +1,57 @@
...
+      template = {
+        spec = {
+          nodeSelector =
+          {{- if hasKeyAny .distribution "nodeSelector" }} {
+            {{- range $key, $value := .distribution.nodeSelector }}
+            "{{ $key }}" = "{{ $value }}"
+            {{- end }}
+          }
+          {{- else }} null
+          {{- end }}
```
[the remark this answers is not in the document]
Al-Pragliola: null here, covers the case where I want to remove the nodeSelector on the manifest, if we use {} instead kubectl patch does nothing

internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
```diff
@@ -186,7 +197,38 @@ func (k *Kubernetes) Exec() error {
 	return nil
 }
 
-func (k *Kubernetes) copyFromTemplate() error {
+func (*Kubernetes) getCommonDataFromDistribution(furyctlCfg template.Config) (map[any]any, []any, error) {
```
omissis, line 200: we could move this code to template.Config perhaps?

configs/provisioners/cluster/eks/coredns.tf.tpl
```diff
@@ -0,0 +1,50 @@
...
+        }
+      }
+    }
+  }
+  coredns_scheduling_patch_as_json = jsonencode(local.coredns_scheduling_patch)
+}
+
+resource "local_file" "cluster_ca" {
+
+  content = base64decode(data.aws_eks_cluster.fury.certificate_authority.0.data)
```
omissis, line 32: nitpick
```suggestion
  content = base64decode(data.aws_eks_cluster.fury.certificate_authority.0.data)
```

configs/provisioners/cluster/eks/coredns.tf.tpl
```diff
@@ -0,0 +1,50 @@
...
+  filename = "${path.module}/secrets/${data.aws_eks_cluster.fury.name}-ca.crt"
+}
+
+resource "null_resource" "patch_coredns" {
+
+  triggers = {
```
omissis, line 38: nitpick
```suggestion
resource "null_resource" "patch_coredns" {
  triggers = {
```

internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
```diff
@@ -246,6 +297,49 @@ func (k *Kubernetes) copyFromTemplate() error {
...
+	}
+
+	reverseMerger := merge.NewMerger(
+		*merger.GetCustom(),
+		*merger.GetBase(),
+	)
+
+	_, err = reverseMerger.Merge()
+	if err != nil {
+		return cfg, fmt.Errorf("error merging files: %w", err)
```
omissis, line 332: we should consider refactoring this code as it's duplicated

omissis (changes requested):

internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
Al-Pragliola: I'm not a big fan, while the input is from this package, the logic inside the function is tied to the schema and may or may not change with other providers/schema versions.

omissis (approved):

configs/provisioners/cluster/eks/coredns.tf.tpl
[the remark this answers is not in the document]
Al-Pragliola: These screenshots above are done PRE Terraform Apply and POST Terraform Apply using the suggested code with {} instead of nil in case of empty keys.

PR #335, opened 2023-03-30
feat: removed sleep usage for ingresses

### LINKED ISSUE #292: Do not use sleep to wait for deletion to complete when "no resouces" are available
Do not use [sleep](https://github.com/sighupio/furyctl/blob/furyctl-ng-alpha1/internal/apis/kfd/v1alpha2/eks/delete/distribution.go#L279) to wait for deletion to complete if there are "no resources" to delete (or better not to use sleep to wait until the task is marked as done).

```
{"level":"info","msg":"Deleting ingresses...","time":"2023-02-21T18:43:16+01:00"}
No resources found
{"level":"info","msg":"Deleting blocking resources...[PersistentVolumeClaims, StatefulSets, Logging, Prometheus]","time":"2023-02-21T18:47:17+01:00"}
No resources found
No resources found
No resources found
No resources found
No resources found
No resources found
...
{"level":"info","msg":"Deleting manifests...","time":"2023-02-21T18:51:26+01:00"}
```

Al-Pragliola:
Changelist:

- added kubernetes client, in order to abstract calls to the cli ( next we may/should create an interface to enable the use of different adapters )
- using aws route53 list-hosted-zones/ list-resource-record-sets to check for pending DNS records instead of "sleep"

fixes #292


internal/x/slices/disjoint.go
```diff
@@ -0,0 +1,51 @@
...
+
+package slices
+
+type TransformFunc[T comparable] func(T) T
+
+func identity[T comparable](v T) T {
+	return v
+}
+
+func Disjoint[T comparable](a, b []T) bool {
```
omissis, line 13: perhaps add a bit of docs to describe what the function does

internal/x/slices/disjoint.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, w := range b {
+		if unique[w] {
+			return false
+		}
+	}
+
+	return true
+}
+
+func DisjointTransform[T comparable](a, b []T, transformA, transformB TransformFunc[T]) bool {
```
omissis, line 29: perhaps add a bit of docs to describe what the function does

internal/kubernetes/client.go
omissis, line 1: it would be wise to add some tests here

internal/kubernetes/client.go
```diff
@@ -0,0 +1,154 @@
...
+func (c *Client) ToolVersion() (string, error) {
+	version, err := c.kubeRunner.Version()
+	if err != nil {
+		return "", fmt.Errorf("error while getting tool version: %w", err)
+	}
+
+	return version, nil
+}
+
+func cleanStringsSlice(slice []string) []string {
```
omissis, line 144: I would move this to the x/slices package and make it generic, replacing the test agains the empty string with a test against the Type's zero value

internal/apis/kfd/v1alpha2/eks/delete/distribution.go
```diff
@@ -324,82 +363,97 @@ func (d *Distribution) deleteBlockingResources() error {
...
+		zones[match[1]] = match[2]
 	}
 
-	logString := log[logStringIndex[0]:logStringIndex[1]]
+	return zones, nil
+}
 
-	if logString != "''" {
-		return fmt.Errorf("%w: %s", errPendingResources, logString)
+func (d *Distribution) checkPendingDNSRecords(hosts []string, hostedZones map[string]string) error {
```
omissis, line 403: I would rename this to something like an assertion, as it fails if the condition is not verified: it is not simply checking it.
```suggestion
func (d *Distribution) assertEmptyDNSRecords(hosts []string, hostedZones map[string]string) error {
```

internal/apis/kfd/v1alpha2/eks/delete/distribution.go
```diff
@@ -324,82 +363,97 @@ func (d *Distribution) deleteBlockingResources() error {
...
-	log, err := d.kubeRunner.Get("all", "ingress", "-o", "jsonpath='{.items[*].metadata.name}'")
+	route53, err := d.awsRunner.Route53("list-hosted-zones", "--query", "HostedZones[*].[Id,Name]",
+		"--output", "text")
 	if err != nil {
-		return fmt.Errorf("error while reading resources from cluster: %w", err)
+		return zones, fmt.Errorf("error getting hosted zones: %w", err)
 	}
 
-	reg := regexp.MustCompile(`'(.*?)'`)
+	re := regexp.MustCompile(`/hostedzone/(\S+)\t(\S+)\.`)
```
omissis, line 392: inefficient use of regex: move it up above as package var to compile it only once

internal/apis/kfd/v1alpha2/eks/delete/distribution.go
```diff
@@ -251,11 +287,20 @@ func (d *Distribution) checkPendingResources() error {
...
+			pvs, errPv = d.kubeClient.GetPersistentVolumes()
+			if errPv == nil && len(pvs) > 0 {
+				errPv = fmt.Errorf("%w: %v", errPendingResources, pvs)
+			}
 
-			errIgrs = d.getIngresses()
+			ingrs, errIgrs = d.kubeClient.GetIngresses()
+			if errIgrs == nil && len(ingrs) > 0 {
+				errIgrs = fmt.Errorf("%w: %v", errPendingResources, ingrs)
+			}
```
omissis, line 303: we could think about running the three calls in parallel to speed up things a bit

internal/apis/kfd/v1alpha2/eks/delete/distribution.go
```diff
@@ -128,57 +141,76 @@ func (d *Distribution) Exec() error {
...
 
 	if err = d.deleteIngresses(); err != nil {
 		return err
 	}
 
+	if len(ingressHosts) > 0 {
+		logrus.Info("Waiting for DNS records to be deleted...")
+
+		err = d.checkPendingDNSRecords(ingressHosts, hostedZones)
+		if err != nil {
```
omissis, line 209:
```suggestion
		if err := d.checkPendingDNSRecords(ingressHosts, hostedZones); err != nil {
```

internal/apis/kfd/v1alpha2/eks/delete/distribution.go
```diff
@@ -324,82 +363,97 @@ func (d *Distribution) deleteBlockingResources() error {
...
+
+		if <-p.C; true {
+			for _, zone := range queue {
+				domains, err := d.awsRunner.Route53("list-resource-record-sets", "--hosted-zone-id", zone,
+					"--query", "ResourceRecordSets[*].Name", "--output", "text")
+				if err != nil {
+					return fmt.Errorf("error getting hosted zone records: %w", err)
+				}
+
+				re := regexp.MustCompile(`(\S+)\.`)
```
omissis, line 435: inefficient use of regex

omissis (changes requested):

internal/kubernetes/client.go
```diff
@@ -0,0 +1,147 @@
...
+	var result []Ingress
+
+	log, err := c.kubeRunner.Get("all", "ingress", "-o",
+		"jsonpath='[{range .items[*]}{\"{\"}\"Name\": \"{.metadata.name}\", "+
+			"\"Host\": [{range .spec.rules[*]}\"{.host}\",{end}]{\"}\"},{end}]'")
+	if err != nil {
+		return result, fmt.Errorf("error while reading resources from cluster: %w", err)
+	}
+
+	logStringIndex := logRegex.FindStringIndex(log)
```
ralgozino, line 63: I did a test with a local cluster with some ingresses, I ran the following command:
```bash
kubectl get ingress -A -o jsonpath='[{range .items[*]}{"{"}"Name": "{.metadata.name}", Host": [{range .spec.rules[*]}"{.host}",{end}]{"}"},{end}]'
```

and the regex is not matching anything... 🤔  see: https://regex101.com/r/51LS6d/1

so I expect that `FindStringIndex(log)` will return `nil` when it should not in this case.

I mean, I guess you tested this and works, so probably I am missing something, for example, I don't see where the `'` that the regex searches would come from. Is it something that the `kubeRunner.Get()` adds?

test/data/e2e/create/cluster/data/kfd.yaml
```diff
@@ -4,18 +4,18 @@
...
-  aws: v2.0.0
-  dr: v1.9.3
-  ingress: v1.12.2
-  logging: v2.0.3
-  monitoring: v1.14.2
-  opa: v1.7.0
+  auth: v0.0.2
+  aws: v2.1.0
+  dr: v1.10.1
+  ingress: feat/change-behaviour-on-externaldns-module
```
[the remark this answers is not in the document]
ralgozino: branch has been merged and is part of the release v1.14.1

internal/apis/kfd/v1alpha2/eks/delete/distribution.go
Al-Pragliola: card created: https://github.com/sighupio/furyctl/issues/341

test/data/e2e/create/cluster/data/kfd.yaml
```diff
@@ -4,18 +4,18 @@
 
 version: v1.24.1
```
ralgozino, line 5: should this be bumped to 1.25 and the modules, installers, etc. to the respective version?

internal/kubernetes/client.go
ralgozino: I agree with @omissis . I'd also suggest adding some comments because the logic in these methods depends heavily on knowing the expected output from the commands that are being run.

ralgozino (changes requested):
Some suggestions and questions:

- I assume it is correct, but just to double-check: this PR merges into `feature/ng-add-support-for-public-eks-clusters` instead of `furyctl-next`, is it OK?

- I see that we have "hard-coded" into the Go logic the objects that need to be removed (i.e. the `deleteBlockingResources()` method). Have you considered this approach vs. having this in the template/distro instead? I'm thinking for example, that if a new resource needs to be added to the list, one would have to remember that this is the place (that "one" usually is someone working on the distro and not on the Go part).

internal/kubernetes/client.go
```diff
@@ -0,0 +1,147 @@
...
+
+	if logStringIndex == nil {
+		return result, nil
+	}
+
+	out := log[logStringIndex[0]+1 : logStringIndex[1]-1]
```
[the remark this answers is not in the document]
Al-Pragliola: yep

internal/kubernetes/client.go
Al-Pragliola: yes, ```jsonpath='[{range .items[*]}{\"{\"}\"Name\": \"{.metadata.name}\", \"Host\": [{range .spec.rules[*]}\"{.host}\",{end}]{\"}\"},{end}]'"```kubeRunner encloses the output of the command between '' so that any output kubectl returns that is not from jsonpath (example "No resource found") won't be parsed by the regex.

internal/kubernetes/client.go
Al-Pragliola: done

test/data/e2e/create/cluster/data/kfd.yaml
Al-Pragliola: yes but I would leave that in a dedicated PR

internal/kubernetes/client.go
```diff
@@ -50,23 +51,25 @@ func NewClient(
...
-	log, err := c.kubeRunner.Get("all", "ingress", "-o",
+	cmdOut, err := c.kubeRunner.Get("all", "ingress", "-o",
 		"jsonpath='[{range .items[*]}{\"{\"}\"Name\": \"{.metadata.name}\", "+
 			"\"Host\": [{range .spec.rules[*]}\"{.host}\",{end}]{\"}\"},{end}]'")
 	if err != nil {
 		return result, fmt.Errorf("error while reading resources from cluster: %w", err)
 	}
 
-	logStringIndex := logRegex.FindStringIndex(log)
+	logStringIndex := cmdOutRegex.FindStringIndex(cmdOut)
```
ralgozino, line 66: I'd suggest renaming `logStringIndex` to something like `singlequoteIndex` to be consistent

ralgozino (changes requested):
one more minor change and I left two questions in my previous review that hasn't been answered yet

test/data/e2e/create/cluster/data/kfd.yaml
ralgozino: ok

internal/kubernetes/client.go
```diff
@@ -0,0 +1,147 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/sighupio/furyctl/internal/tool/kubectl"
+	execx "github.com/sighupio/furyctl/internal/x/exec"
+	"github.com/sighupio/furyctl/internal/x/slices"
+)
+
+var logRegex = regexp.MustCompile(`'(.*?)'`)
```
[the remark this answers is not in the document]
Al-Pragliola: done

Al-Pragliola:
> Some suggestions and questions:
> 
> * I assume it is correct, but just to double-check: this PR merges into `feature/ng-add-support-for-public-eks-clusters` instead of `furyctl-next`, is it OK?
> * I see that we have "hard-coded" into the Go logic the objects that need to be removed (i.e. the `deleteBlockingResources()` method). Have you considered this approach vs. having this in the template/distro instead? I'm thinking for example, that if a new resource needs to be added to the list, one would have to remember that this is the place (that "one" usually is someone working on the distro and not on the Go part).

- Yes, in order to minimize conflicts, I started from that branch.
- agreed, we should discuss more in detail about orphaned resources in general to find a way more robust/scalable

PR #336, opened 2023-03-31
feat: implement .fctlstate file

### LINKED ISSUE #216: Attention point using the Home Directory as base for bin and cluster stuffs
Attention points:
- as all binaries by defautl are downloaded into `~/.furyctl/bin`, it's not possible to download multiple distribution dependencies in parallel without raise errors.
- if I have multiple AWS accounts and in each ones I use the same cluster name, I can't execute parallel operations without the risk to override the downloaded stuffs.

IMHO, I suggest to use the a more unique ID as `apiVersion/Kind/AWS_ID/version/cluster Name`

Proposed solution:

- we should introduce a lockfile (and a mechanism to remove it in case the command is terminated abruptly) to avoid having parallel process overwrite the same file(s)
- we should store the cluster folders in the `~/.furyctl/clusters` folder (instead of the `~/.furyctl` one as we are doing now)
- we should use the following naming schema for the cluster folders: `<cluster-name>-<region>-<kind>-<cloud-provider-id>`. eg: `eks-barebone-eu-west-1-EKSCluster-492816857163`

alessiodionisi:
Partially closes #216.

## Changes
- Implemented `.fctlstate` state file with JSON encoding
- Moved clusters generated files to `~/.furyctl/clusters/<GUID>`
- Used the state file to save the generated GUID


alessiodionisi:
rebased

alessiodionisi:
rebased and squashed

alessiodionisi:
also runned `go mod tidy`

omissis:
closing as not relevant for now, we'll pick it up in the future if needed

PR #353, opened 2023-04-13
fix: don't ask VPN connection if cluster is public or public/private

### LINKED ISSUE #349: Furyctl (next) - Furyctl asks for `enter` even if i'm creating a public cluster
As per title, furyctl asks for `enter` even if i'm creating a public cluster

alessiodionisi:
Closes #349.


omissis (approved):

Al-Pragliola (changes requested):
lint

PR #355, opened 2023-04-13
feat: log details when deleting blocking resources

### LINKED ISSUE #351: Improve cluster deletion messages
When deleting a cluster, you get the following message:

```yaml
INFO Deleting blocking resources...[PersistentVolumeClaims, StatefulSets, Logging, Prometheus]
```

This is a time-consuming operation and there’s almost no feedback to the user for several minutes.

furyctl should warn the user that the operation will take some time or even better, if we can, show some progress. For example, instead of only one message for all of PVC, STS, Logging and Prometheus print one message for each one.

alessiodionisi:
Closes #351.


omissis (changes requested):
Please add a warning for the user, stating the operation will take some time.

PR #359, opened 2023-04-14
feature/check resources before deleting them

omissis:
- feat: add existence check on single resources before deleting them
- chore: bump fury-distribution dep
- fix: add kubectl get flag that allows for existance check without errors


Al-Pragliola (approved):
lgtm

Issue #313, opened 2023-03-01
Missing check on EKS cluster when applying changes


g-iannelli:
When apply the changes on EKS cluster, there is no check about the cluster replacement.

```shell
pgrep openvpn | xargs sudo kill -9 && furyctl-ng create cluster -c barebone.eu-west-1-alphav2.yml --distro-location /Users/giuseppe.iannelli/Workspace/github.com/sighupio/fury-distribution --skip-deps-download
INFO Downloading distribution...
INFO Validating configuration file...
INFO Validating dependencies...
INFO Creating cluster...
INFO Creating infrastructure...
INFO Creating cloud resources, this could take a while...
INFO Connecting to VPN, you will be asked for your SUDO password...
INFO Creating Kubernetes Fury cluster...
INFO Checking connection to the VPC...
INFO Creating cloud resources, this could take a while...
ERRO error while creating cluster: error while executing kubernetes phase: cannot create cloud resources: cannot create cloud resources: /Users/giuseppe.iannelli/.furyctl/bin/terraform/0.15.4/terraform /Users/giuseppe.iannelli/.furyctl/bin/terraform/0.15.4/terraform apply -no-color -json plan/terraform.plan: command failed - exit status 1
out: {"@level":"info","@message":"Terraform 0.15.4","@module":"terraform.ui","@timestamp":"2023-03-01T11:03:22.648354+01:00","terraform":"0.15.4","type":"version","ui":"0.1.0"}
{"@level":"info","@message":"module.fury.module.cluster.aws_eks_cluster.this[0]: Plan to replace","@module":"terraform.ui","@timestamp":"2023-03-01T11:03:27.200927+01:00","change":{"resource":{"addr":"module.fury.module.cluster.aws_eks_cluster.this[0]","module":"module.fury.module.cluster","resource":"aws_eks_cluster.this[0]","implied_provider":"aws","resource_type":"aws_eks_cluster","resource_name":"this","resource_key":0},"action":"replace","reason":"cannot_update"},"type":"planned_change"}
[...]

PR #374, opened 2023-05-02
Feat: prompt on critical resource delete

Al-Pragliola:
Changelist:

- added terraform plan parser to detect changes: destroy, replace, change, add
- prompt user when trying to delete critical resources after plan in infrastructure and kubernetes phase during cluster creation
- intersection function in x/slices

List of critical resources checked for each phase:

Infrastructure: 
- "aws_vpc"
- "aws_subnet"

Kubernetes:
- "aws_eks_cluster"

closes: #313

### DIFF
--- a/internal/apis/kfd/v1alpha2/eks/create/infrastructure.go
+++ b/internal/apis/kfd/v1alpha2/eks/create/infrastructure.go
@@ -132,7 +132,12 @@ func (i *Infrastructure) Exec() error {
 
 		prompter := iox.NewPrompter(bufio.NewReader(os.Stdin))
 
-		if !prompter.Ask("yes") {
+		prompt, err := prompter.Ask("yes")
+		if err != nil {
+			return fmt.Errorf("error reading user input: %w", err)
+		}
+
+		if !prompt {
 			return ErrAbortedByUser
 		}
 	}
--- a/internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
+++ b/internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
@@ -171,7 +171,12 @@ func (k *Kubernetes) Exec() error {
 
 		prompter := iox.NewPrompter(bufio.NewReader(os.Stdin))
 
-		if !prompter.Ask("yes") {
+		prompt, err := prompter.Ask("yes")
+		if err != nil {
+			return fmt.Errorf("error reading user input: %w", err)
+		}
+
+		if !prompt {
 			return ErrAbortedByUser
 		}
 	}
--- a/internal/x/io/prompter.go
+++ b/internal/x/io/prompter.go
@@ -6,6 +6,7 @@ package iox
 
 import (
 	"bufio"
+	"fmt"
 	"strings"
 )
 
@@ -19,10 +20,10 @@ func NewPrompter(r *bufio.Reader) *Prompter {
 	}
 }
 
-func (p *Prompter) Ask(w string) bool {
+func (p *Prompter) Ask(w string) (bool, error) {
 	response, err := p.Reader.ReadString('\n')
 	if err != nil {
-		return false
+		return false, fmt.Errorf("error reading from stdin: %w", err)
 	}
 
 	response = strings.TrimSuffix(response, "\n")
@@ -31,5 +32,5 @@ func (p *Prompter) Ask(w string) bool {
 	return strings.Compare(
 		strings.ToLower(response),
 		strings.ToLower(w),
-	) == 0
+	) == 0, nil
 }
--- a/cmd/delete/cluster.go
+++ b/cmd/delete/cluster.go
@@ -181,7 +181,12 @@ func NewClusterCmd(tracker *analytics.Tracker) *cobra.Command {
 
 				prompter := iox.NewPrompter(bufio.NewReader(os.Stdin))
 
-				if !prompter.Ask("yes") {
+				prompt, err := prompter.Ask("yes")
+				if err != nil {
+					return fmt.Errorf("error reading user input: %w", err)
+				}
+
+				if !prompt {
 					return nil
 				}
 			}
--- a/internal/x/io/prompter_test.go
+++ b/internal/x/io/prompter_test.go
@@ -61,7 +61,12 @@ func TestPrompter_Ask(t *testing.T) {
 
 			prompter := iox.NewPrompter(bufio.NewReader(strings.NewReader(tc.input)))
 
-			if prompter.Ask(tc.promptWord) != tc.expected {
+			prompt, err := prompter.Ask(tc.promptWord)
+			if err != nil {
+				t.Errorf("unexpected error: %v", err)
+			}
+
+			if prompt != tc.expected {
 				t.Errorf("expected %v, got %v", tc.expected, !tc.expected)
 			}
 		})


internal/apis/kfd/v1alpha2/eks/create/infrastructure.go
```diff
@@ -103,14 +108,35 @@ func (i *Infrastructure) Exec() error {
...
 
+	tfParser := parser.NewTfPlanParser(plan)
+
+	parsedPlan := tfParser.Parse()
+
+	eksInf := eks.NewInfra()
+
+	criticalResources := slices.Intersection(eksInf.GetCriticalTFResources(), parsedPlan.Destroy)
+
+	if len(criticalResources) > 0 {
```
ralgozino, line 128: IIRC we have a `--force` flag that makes furyctl not ask for confirmation so we can run the delete command with automation (a script for example), maybe we should have something similar here?

A `create` operation that is destroying critical resources should not be taken lightly. So at first glance, I'd say we shouldn't add a flag to skip this check, but what do you think? Have you thought about this?

internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
```diff
@@ -143,14 +147,35 @@ func (k *Kubernetes) Exec() error {
...
+	parsedPlan := tfParser.Parse()
+
+	eksKube := eks.NewKubernetes()
+
+	criticalResources := slices.Intersection(eksKube.GetCriticalTFResources(), parsedPlan.Destroy)
+
+	if len(criticalResources) > 0 {
+		logrus.Warnf("A deletion of critical resources has been detected: %s. Please check the plan carefully.",
+			strings.Join(criticalResources, ", "))
+		logrus.Warn("When you are ready to proceed, write 'yes' to continue or anything else to abort: ")
```
ralgozino, line 170: same as above:
```suggestion
		logrus.Warnf("Deletion of the following critical resources has been detected: %s. See the logs for more details.",
			strings.Join(criticalResources, ", "))
		logrus.Warn("Do you want to proceed? write 'yes' to continue or anything else to abort: ")
```

ralgozino (changes requested):
I left some suggestions.

Even though this PR is a nice addition, I don't think the PR fixes the related issue, for example, if a cluster already exists with the same name but it is not in the terraform state, terraform won't try to delete it so the check would not find it. The user will get an error like in the related issue. Am I right?

internal/eks/kubernetes.go
```diff
@@ -0,0 +1,19 @@
+// Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package eks
+
+type Kubernetes struct {
+	criticalTFResources []string
```
g-iannelli, line 8:
```suggestion
	criticalTFResourceTypes []string
```

internal/eks/kubernetes.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+func NewKubernetes() *Kubernetes {
+	return &Kubernetes{
+		criticalTFResources: []string{"aws_eks_cluster"},
```
g-iannelli, line 13:
```suggestion
		criticalTFResourceTypes: []string{"aws_eks_cluster"},
```

internal/eks/kubernetes.go
```diff
@@ -0,0 +1,19 @@
...
+	}
+}
+
+func (k *Kubernetes) GetCriticalTFResources() []string {
+	return k.criticalTFResources
```
g-iannelli, line 18:
```suggestion
func (k *Kubernetes) GetCriticalTFResourceTypes() []string {
	return k.criticalTFResourceTypes
```

internal/eks/infra.go
```diff
@@ -0,0 +1,19 @@
...
+}
+
+func NewInfra() *Infra {
+	return &Infra{
+		criticalTFResources: []string{"aws_vpc", "aws_subnet"},
+	}
+}
+
+func (i *Infra) GetCriticalTFResources() []string {
+	return i.criticalTFResources
```
g-iannelli, line 18:
```suggestion
func (i *Infra) GetCriticalTFResourceTypes() []string {
	return i.criticalTFResourceTypes
```

internal/apis/kfd/v1alpha2/eks/create/kubernetes.go
```diff
@@ -143,14 +147,35 @@ func (k *Kubernetes) Exec() error {
...
 		return nil
 	}
 
+	tfParser := parser.NewTfPlanParser(plan)
+
+	parsedPlan := tfParser.Parse()
+
+	eksKube := eks.NewKubernetes()
+
+	criticalResources := slices.Intersection(eksKube.GetCriticalTFResources(), parsedPlan.Destroy)
```
g-iannelli, line 165:
```suggestion
	criticalResources := slices.Intersection(eksKube.GetCriticalTFResourceTypes(), parsedPlan.Destroy)
```

internal/apis/kfd/v1alpha2/eks/create/infrastructure.go
```diff
@@ -103,14 +108,35 @@ func (i *Infrastructure) Exec() error {
...
 		return nil
 	}
 
+	tfParser := parser.NewTfPlanParser(plan)
+
+	parsedPlan := tfParser.Parse()
+
+	eksInf := eks.NewInfra()
+
+	criticalResources := slices.Intersection(eksInf.GetCriticalTFResources(), parsedPlan.Destroy)
```
g-iannelli, line 126:
```suggestion
	criticalResources := slices.Intersection(eksInf.GetCriticalTFResourceTypes(), parsedPlan.Destroy)
```

g-iannelli (changes requested):

internal/apis/kfd/v1alpha2/eks/create/infrastructure.go
Al-Pragliola: yep, I think making --force also skip this check would be worse than lock a script, maybe in the future we can think of a "--ci" flag to instead return an error instead of prompting the user

internal/parser/terraform.go
```diff
@@ -0,0 +1,68 @@
...
+			pl.Change = append(pl.Change, resourceName)
+
+		case "-/+":
+			pl.Destroy = append(pl.Destroy, resourceName)
+			pl.Add = append(pl.Add, resourceName)
+		}
+	}
+
+	return &pl
+}
```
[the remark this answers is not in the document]
Al-Pragliola: Fair enough, we can consider to rework this using terraform-json library and terraform show -json

https://github.com/sighupio/furyctl/issues/375

internal/parser/terraform.go
```diff
@@ -0,0 +1,68 @@
...
+		case "-":
+			pl.Destroy = append(pl.Destroy, resourceName)
+
+		case "~":
+			pl.Change = append(pl.Change, resourceName)
+
+		case "-/+":
+			pl.Destroy = append(pl.Destroy, resourceName)
+			pl.Add = append(pl.Add, resourceName)
+		}
```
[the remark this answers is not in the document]
Al-Pragliola: added

internal/tool/terraform/runner.go
```diff
@@ -67,8 +67,9 @@ func (r *Runner) Init() error {
 	return nil
 }
 
-func (r *Runner) Plan(timestamp int64, params ...string) error {
+func (r *Runner) Plan(timestamp int64, params ...string) (string, error) {
 	args := []string{"plan"}
+	out := ""
```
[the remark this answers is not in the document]
g-iannelli: #375 should eliminate the need for the return `out` or am I wrong?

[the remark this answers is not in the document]
Al-Pragliola: I would still split the function in two separate ones Plan() and SavePlanToDisk() or something like that

ralgozino (approved):
lgtm

internal/x/io/prompter.go
```diff
@@ -0,0 +1,35 @@
...
+func NewPrompter(r *bufio.Reader) *Prompter {
+	return &Prompter{
+		Reader: r,
+	}
+}
+
+func (p *Prompter) Ask(w string) bool {
+	response, err := p.Reader.ReadString('\n')
+	if err != nil {
+		return false
```
omissis, line 25: it would be better not to hide the errors: if it fails here we could have a fairly hard time understanding where it broke.

omissis (approved):

PR #381, opened 2023-05-15
feat: validate deps checksums before downloading

### LINKED ISSUE #181: Dependencies checksum
It could happen that a download for a dependency gets corrupted. furyctl-ng shall check that the binary matches a predefined checksum, this could also be used to know if it needs to download the dependency or the already present one is correct and skip the downloading.

alessiodionisi:
Closes #181.

SSIA


omissis:
Closing as we are not 100% convinced about the implementation. To be picked up again in the future.

Issue #143, opened 2023-01-03
Refactor tfvars generation


ralgozino:
refactor how we dump the `tfvars` file in furyctl, possibly using go templates instead of string concats

ralgozino:
I believe this has been done.

PR #383, opened 2023-05-15
feature/upgrade terraform to 1.x

omissis:
- update codebase to make it work with terraform 1.x modules
- fix: add output fake command for tf runner tests

### DIFF
--- a/cmd/connect/openvpn.go
+++ b/cmd/connect/openvpn.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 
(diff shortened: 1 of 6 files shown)


Al-Pragliola (approved):
Probably we should split the command Apply in two Apply and Output but there are many things to refactor and we should probably address it when working on https://github.com/sighupio/furyctl/issues/143

omissis:
@Al-Pragliola you're right, I'm refactoring it within this PR.

PR #399, opened 2023-06-19
feat: add yq dependency, support apply script from manifest templates

### LINKED ISSUE #393: Evaluate what to do to enhance kubectl apply stage on Furyctl distribution phase
Define what is the best approach to apply the distribution on a target cluster. Right now, we are simply kubectl applying multiple times hoping that it's enough.

One approach that is working without errors is the following:

```
#!/bin/bash

kustomizebin=../../../bin/kustomize/3.10.0/kustomize
kubectlbin=../../../bin/kubectl/1.25.8/kubectl
yqbin=yq

$kustomizebin build . | $yqbin 'select(.kind == "CustomResourceDefinition")' | $kubectlbin apply -f - --server-side
$kustomizebin build . | $yqbin 'select(.kind == "CustomResourceDefinition")' | $kubectlbin wait --for condition=established --timeout=60s -f -

$kustomizebin build . \
  | $yqbin 'select(.kind != "Issuer" and .kind != "ClusterIssuer" and .kind != "Certificate" and .kind != "Ingress" and .kind != "K8sLivenessProbe" and .kind != "K8sReadinessProbe" and .kind != "K8sUniqueIngressHost" and .kind != "SecurityControls")' \
  | $kubectlbin apply -f - --server-side   

$kustomizebin build . | $yqbin 'select(.kind == "Deployment" and .metadata.namespace =="cert-manager" )' | $kubectlbin wait --for condition=available --timeout=120s -f -

$kustomizebin build . \
  | $yqbin 'select(.kind == "Issuer" or .kind == "ClusterIssuer" or .kind == "Certificate")' \
  | $kubectlbin apply -f - --server-side

$kubectlbin get pods -o yaml -n ingress-nginx | $kubectlbin wait --for condition=Ready --timeout=180s -f -

$kustomizebin build . | $kubectlbin apply -f - --server-side
```

We need to check if there are some alternatives using kustomize or other tools that achieve the same, or make this script (and the deletion script) part of the fury-distribution templates. A template is useful when, for example, there is no ingress controller in the cluster (because we are disabling it) and we want to skip the step for example: `$kubectlbin get pods -o yaml -n ingress-nginx | $kubectlbin wait --for condition=Ready --timeout=180s -f -`

PS: binaries should be input passed as argument by furyctl (since we are using our own binary in the path ~/.furyctl/bin)
```[tasklist]
### Tasks
```

alessiodionisi:
Closes #393.

⚠️ Merge after https://github.com/sighupio/fury-distribution/pull/107 and update `go.mod` dependency


internal/apis/kfd/v1alpha2/distribution/delete/distribution.go
```diff
@@ -93,239 +93,47 @@ func (d *Distribution) Exec() error {
...
-	logrus.Info("Checking pending resources...")
-
-	return d.checkPendingResources()
-}
-
-func (d *Distribution) buildManifests() (string, error) {
 	kzOut, err := d.kzRunner.Build()
 	if err != nil {
-		return "", fmt.Errorf("error building manifests: %w", err)
+		return fmt.Errorf("error building manifests: %w", err)
```
Al-Pragliola, line 101: Do we need to run kustomize build, given that now we do that in delete script?

Al-Pragliola (changes requested):

PR #409, opened 2023-07-06
Feat: update installer to fix spot with launch templates

nutellinoit:
As per title, this updates the installer on the legacy version and removes the brew step on goreleaser


Al-Pragliola (approved):
LGTM, if we are not going to use brew anymore on legacy we can also delete the lines instead of commenting

PR #430, opened 2023-09-19
feat: path URLs relative to furyctl.yaml, prepare for 1.26 release

### LINKED ISSUE #429: Path URLs relative to furyctl.yaml
Path URLs that starts with `./` should be relative to `furyctl.yaml`

alessiodionisi:
Closes #429.

- Add path URLs relative to furyctl.yaml

The path should start with `./`.


internal/template/mapper/mapper.go
```diff
@@ -119,6 +121,17 @@ func injectDynamicResString(val string) (string, error) {
 		val = strings.Replace(val, dynamicValue, parsedDynamicValue, 1)
 	}
 
+	// If the value is a relative path, we need to convert it to an absolute path.
+	isPath := regexp.MustCompile(`^.\/(.*)$`).MatchString(val)
```
omissis, line 125: this can be optimized by moving the MustCompile to a package variable, to avoid the compilation to happen over and over again

internal/template/mapper/mapper.go
```diff
@@ -108,7 +109,8 @@ func injectDynamicRes(
 	return m, nil
 }
 
-func injectDynamicResString(val string) (string, error) {
+func mapDynamicValuesAndPathsString(val string) (string, error) {
+	// If the value contains dynamic values, we need to parse them.
 	dynamicValues := regexp.MustCompile(`{(.*?)}`).FindAllString(val, -1)
```
omissis, line 114: this can be optimized by moving the MustCompile to a package variable, to avoid the compilation to happen over and over again

nutellinoit (approved):
LGTM, tested EKS and Distro only, everything is working

PR #439, opened 2023-10-16
feat: diff state

Al-Pragliola:


internal/apis/kfd/v1alpha2/distribution/rules/builder.go
```diff
@@ -0,0 +1,69 @@
...
+type DistroRulesSpec struct {
+	Distribution []Rule `yaml:"distribution"`
+}
+
+type Rule struct {
+	Path      string `yaml:"path"`
+	Immutable bool   `yaml:"immutable"`
+}
+
+type Builder interface {
```
omissis, line 26: this is looks a bit more like a factory than a builder, but I would probably call it Extractor or ImmutablesExtractor?

internal/apis/kfd/v1alpha2/distribution/rules/builder.go
```diff
@@ -0,0 +1,69 @@
...
+	GetImmutables(phase string) []string
+}
+
+type DistroBuilder struct {
+	Spec DistroRulesSpec
+}
+
+func NewDistroClusterRulesBuilder(distributionPath string) (*DistroBuilder, error) {
```
omissis, line 34: following up the comment above, this is probably a NewDistroImmutablesExtractor

internal/apis/kfd/v1alpha2/onpremises/create/preflight.go
```diff
@@ -0,0 +1,292 @@
...
+package create
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path"
+	"path/filepath"
+
+	diffx "github.com/r3labs/diff/v3"
```
omissis, line 14: we usually use `x` as a suffix for our internal "x" libraries. how about `r3diff`

omissis (changes requested):

Al-Pragliola (commented):
Falling back to Draft, the preflight phase is not working correctly with EKS private clusters because it's missing the ovpn certificate regen (idempotency)

omissis (dismissed):

omissis (approved):

PR #448, opened 2023-11-13
Introduce terraform state bucket creation in furyctl create cluster

SIGHUP-C-3PO:
- feat: autocreate the terraform state bucket before starting the apply action

### DIFF
--- a/internal/apis/kfd/v1alpha2/eks/create/preflight.go
+++ b/internal/apis/kfd/v1alpha2/eks/create/preflight.go
@@ -38,7 +38,7 @@ import (
 const vpnDefaultPort = 1194
 
 var (
-	ErrAWSS3BucketNotFound                   = errors.New("AWS S3 Bucket not found, please create it before running furyctl")
+	ErrAWSS3BucketNotFound                   = errors.New("AWS S3 Bucket not found")
 	ErrAWSS3BucketRegionMismatch             = errors.New("AWS S3 Bucket region mismatch")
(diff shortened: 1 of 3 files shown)


alessiodionisi (approved):

nutellinoit:
What happens if the bucket already exists?

internal/apis/kfd/v1alpha2/eks/create/preflight.go
```diff
@@ -549,3 +560,68 @@ func (p *PreFlight) CheckReducerDiffs(d r3diff.Changelog, diffChecker diffs.Chec
...
+		"--region", region,
+		"--create-bucket-configuration", fmt.Sprintf("LocationConstraint=%s", region),
+	); err != nil {
+		return fmt.Errorf("%w: %w", ErrCannotCreateTerraformStateAWSS3Bucket, err)
+	}
+
+	return nil
+}
+
+func (p *PreFlight) getTerraformStateAWSS3Bucket() error {
```
Al-Pragliola, line 598: nit: assertTerraformStateAWSS3BucketMatches or similar instead of get to be consistent with other function names.

omissis:
@nutellinoit nothing, it only creates it if it's not there

internal/apis/kfd/v1alpha2/eks/create/preflight.go
omissis: pushed to develop

PR #451, opened 2023-11-27
feat: cluster upgrade state

Al-Pragliola:


internal/upgrade/store.go
```diff
@@ -0,0 +1,183 @@
...
+		"Kubernetes":         cluster.OperationPhaseKubernetes,
+		"PostKubernetes":     cluster.OperationSubPhasePostKubernetes,
+		"PreDistribution":    cluster.OperationSubPhasePreDistribution,
+		"Distribution":       cluster.OperationPhaseDistribution,
+		"PostDistribution":   cluster.OperationSubPhasePostDistribution,
+		"":                   cluster.OperationPhaseAll,
+	}
+
+	return mapSubPhase[phase]
+}
```
omissis, line 183: perhaps these belong to the cluster package?

internal/cluster/phase.go
```diff
@@ -17,38 +17,66 @@ import (
...
-var errUnsupportedPhase = errors.New(
-	"unsupported phase, options are: infrastructure, kubernetes, distribution, plugins",
+var (
+	errUnsupportedPhase = errors.New(
+		"unsupported phase, options are: infrastructure, kubernetes, distribution, plugins",
+	)
+	errUnsupportedOperationPhase = errors.New(
+		"unsupported operation phase, options are: pre-infrastructure, infrastructure, post-infrastructure, " +
+			"pre-kubernetes, kubernetes, post-kubernetes, pre-distribution, distribution, post-distribution, plugins",
+	)
```
omissis, line 44: These errors are returned by exported functions (hence they are visible from other packages) and therefore participate in defining the package's api: this means it would be probably better to export them, too, to let client code handle them should the need arise.

internal/cluster/phase.go
```diff
@@ -188,14 +216,18 @@ func (*OperationPhase) CopyFromTemplate(
 	return nil
 }
 
-func (p *OperationPhase) CopyPathsToConfig(cfg *template.Config) {
+func (cp *OperationPhase) CopyPathsToConfig(cfg *template.Config) {
```
omissis, line 219: nitpick
```suggestion
func (op *OperationPhase) CopyPathsToConfig(cfg *template.Config) {
```

internal/cluster/phase.go
```diff
@@ -188,14 +216,18 @@ func (*OperationPhase) CopyFromTemplate(
...
+		"helmfile":   cp.HelmfilePath,
+		"kubectl":    cp.KubectlPath,
+		"kustomize":  cp.KustomizePath,
+		"terraform":  cp.TerraformPath,
+		"vendorPath": path.Join(cp.Path, "..", "vendor"),
+		"yq":         cp.YqPath,
 	}
 }
+
+func (cp *OperationPhase) Self() *OperationPhase {
```
omissis, line 231: nitpick
```suggestion
func (op *OperationPhase) Self() *OperationPhase {
```

internal/cluster/phase.go
omissis: same for the other functions above

PR #471, opened 2024-01-31
Feature/upgrade improvements

Al-Pragliola:
- added `--skip-nodes-upgrade` flag to upgrade just the control plane nodes
- added `--upgrade-node` flag to select and upgrade a single worker node
- added `--pod-running-check-timeout` flag to specify a timeout for the pod check running after the worker nodes upgrade


omissis (approved):
LGTM

PR #475, opened 2024-02-06
optimize the behaviour on downloading dependencies

omissis:
- chore: make GitProtocol attribute on the Downloader struct not exposed as there was no need to have it public
- feat: introduce dependencies cache
- feat: introduce distribution cache
- fix: work around issues caused to cache by tools renamings
- feat: introduce concurrent dependencies downloads
- feat: introduce concurrent tools and modules downloads


Al-Pragliola (approved):
lgtm

PR #477, opened 2024-02-08
normalize providers folder names

omissis:
- chore: rename eks and distribution folders in apis folder to match their provider's names
- chore: refactor internal/eks folder


Al-Pragliola (approved):
lgtm

PR #478, opened 2024-02-09
feat(eks-installer):  integrate the v3.1.0 eks installer to enable logstype

smerlos:
This PR integrates EKS Installer version 3.1.0, introducing enhancements to Kubernetes logging configuration within Terraform projects. Key features include:

- **Log Type Configuration:** Adds capability to specify Kubernetes log types in the config file via `.spec.kubernetes.logsType`
- **Value Validation:** The CLI now checks for the presence of `.spec.kubernetes.logsType` and validates against allowed EKS values, including support for an empty array (`[]`) for flexibility.
- **Backward Compatibility:** Maintains previous behavior in the absence of new configurations, ensuring compatibility with earlier project versions.

### Testing:

- **Configuration Validation:** Tests with valid and invalid `.spec.kubernetes.logsType` values confirmed correct handling.
- **With/Without Configuration:** The installer's behavior was tested both with and without the new `.spec.kubernetes.logsType` field, ensuring proper functionality.
- **Backward Compatibility:** Changes were verified to be backward compatible, defaulting to previous settings when new log configurations are not specified.


Al-Pragliola (approved):
LGTM

PR #481, opened 2024-02-14
feat: force flag options

alessiodionisi:
Change `--force` from `bool` to `[]string` to support options.

Options available:
- all
- migrations
- upgrades
- pods-running-check


Al-Pragliola (approved):
lgtm

PR #482, opened 2024-02-14
feat: update to new KFD release

Al-Pragliola:


nutellinoit (dismissed):
LGTM, we should test it after the release of the new KFD versions

alessiodionisi (approved):

PR #498, opened 2024-03-29
Add support to relative paths with {path://} dynamic value

alessiodionisi:
This feature ad a new "dynamic value" for string interpolation the name `path` that converts a relative path to absolute.

NOTE: This feature can replace the string regex replacer that converts any string that starts with `./` as absolute.

## Usage example
Secret generator with files:

```yaml
customPatches:
  secretGenerator:
    - name: my-secret
      namespace: default
      type: kubernetes.io/tls
      files:
        - tls.crt={path://./certs/tls.crt}
        - tls.key={path://./certs/tls.key}
```

### DIFF
--- a/internal/parser/config.go
+++ b/internal/parser/config.go
@@ -15,6 +15,7 @@ import (
 )
 
 const (
+	Path  = "path"
 	Env   = "env"
 	File  = "file"
 	HTTP  = "http"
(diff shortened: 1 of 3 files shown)


Al-Pragliola (approved):
LGTM

PR #500, opened 2024-04-18
feat: add upgrade paths for 1.29

nutellinoit:
- [x] Add upgrade paths
- [x] Make furyctl compatible with 1.29.0

Note: there is no need to update the fury-distribution library, there are no changes in the schema


Al-Pragliola (approved):
lgtm

PR #520, opened 2024-06-27
feat: use viper to map env vars as flag

Al-Pragliola:
- fixed viper to make it possible to use env vars with prefix FURYCTL_ (eg. FURYCTL_DRY_RUN, FURYCTL_DISTRO_LOCATION)
- refactored code to remove dependencies in cmds with a service container


alessiodionisi (dismissed):

ralgozino (dismissed):
The PR looks good to me and works as intended in my tests 🎉 

I tested setting the following env vars:

```
FURYCTL_OUTDIR = '~/src/test/furyctl/'
FURYCTL_DISTRO_LOCATION = '~/src/github.com/sighupio/fury-distribution'
FURYCTL_DISABLE_ANALYTICS = "true"
FURYCTL_DEBUG = "true"
FURYCTL_SKIP_DEPS_DOWNLOAD = "true"
```

and the flags where set correspondingly as expected 😃 

This makes working with furyctl+direnv (or mise in my case) a breeze 💪 

For a normal user though, the env vars are not discoverable. Can we add at leas some mention to them in the readme, furyctl docs, or maybe in the help messages?

I'm approving the PR anyway because the docs can be done separately.

PR #525, opened 2024-07-18
feat(create pki): add command create pki to generate PKI files needed for on-prem clusters

ralgozino:
This is a port of a piece of functionality currently present on furyagent. The code has been modified and simplified to cover only the PKI creation case, original code on furyagent supported much more options like backups and automatically joining nodes in AWS.

The PR adds a `create pki` command with options to create the Public Key Infrastructure needed for on-premises Kubernetes clusters (etcd CA and certs, Kubernetes control plane CA and certs). 

The command allows limiting which PKI to create (etcd only, control plane only or both) and also in which path to save the files.

The resulting files where tested creating a cluster from scratch with the PKI generated with this command and all is working with the files created by furyagent.

Notice that this PR adds some `replace` directives to the `go.mod` file to use some libraries from Kubernetes (as was done in furyagent), this new dependencies increase the binary size significantly (~30MB).

I also updated the main readme to notice the user of this new feature and fix some wording. I did not write any tests though.


Al-Pragliola (approved):
lgtm

PR #529, opened 2024-08-01
chore(lockfile): improve error message

ralgozino:
make the error message when the lock file exists more user friendly. Explaining what does it mean when the file already exists.

Before:
image

After:
image


Al-Pragliola (approved):
lgtm

PR #544, opened 2024-10-03
fix: fix outdir and workdir on commands

alessiodionisi:
Fixed `get kubeconfig`, `get upgrade-paths` and `renew certificates` commands.


alessiodionisi:
@ralgozino you could take a test of `get kubeconfig`, `get upgrade-paths` and `renew certificates` command please? thanks

ralgozino (approved):
The 3 commands now respect the outdir when passed and do not use the $HOME/.furyctl folder anymore 🎉 

I also tested setting `--outdir` and `--workdir` at the same time and everything seems to be working as expected.

well done!

PR #546, opened 2024-10-08
prepare for release 0.29.9

alessiodionisi:
- Prepared for release v0.29.9
- Upgraded golang
- Upgrade dependecies


go.mod
```diff
@@ -1,172 +1,168 @@
 module github.com/sighupio/furyctl
 
-go 1.22
-
-toolchain go1.22.3
+go 1.23.2
```
ralgozino, line 3: do we want to pin the patch version too? Or would it be better to use `1.23` instead of `1.23.2` like we did for `1.22`?

alessiodionisi: `go` with patch version is the replacement of the old `toolchain`

ralgozino (approved):

PR #548, opened 2024-10-14
fix: set clusterissuer route53 zone id by reading terraform outputs

alessiodionisi:
## Changes

- Added missing `spec.distribution.modules.ingress.certManager.clusterIssuer.route53.hostedZoneId` on the EKS private schema by reading it from the Terraform outputs.
- Region is always eu-west-1 (is this ok?).

Closes #https://github.com/sighupio/fury-distribution/issues/134.

### DIFF
--- a/internal/apis/kfd/v1alpha2/ekscluster/common/distribution.go
+++ b/internal/apis/kfd/v1alpha2/ekscluster/common/distribution.go
@@ -30,6 +30,7 @@ var (
 	errCastingDNSPubIamToStr = errors.New("error casting external_dns_public_iam_role_arn output to string")
 	errCastingCertIamToStr   = errors.New("error casting cert_manager_iam_role_arn output to string")
 	errCastingVelIamToStr    = errors.New("error casting velero_iam_role_arn output to string")
+	errTerraformOutputToStr  = errors.New("error casting terraform output to string")
 )
 
 type Distribution struct {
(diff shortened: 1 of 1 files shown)


internal/apis/kfd/v1alpha2/ekscluster/common/distribution.go
```diff
@@ -353,5 +355,13 @@ func (d *Distribution) extractARNsFromTfOut() (map[string]string, error) {
 		}
 	}
 
+	awsRoute53ZonePublicID, ok := out["aws_route53_zone_public_id"]
+	if ok {
+		arns["aws_route53_zone_public_id"], ok = awsRoute53ZonePublicID.Value.(string)
+		if !ok {
+			return nil, errTerraformOutputToStr
```
ralgozino, line 362: if we are returning this error only in this case, you should be more specific in the error text, saying that the Zone could not be found or something like that intead of error casting terraform output to string.

ralgozino (changes requested):
Just a nit pick on the error handling. Besides that code-wise looks good, have you tested it?

alessiodionisi:
> Just a nit pick on the error handling. Besides that code-wise looks good, have you tested it?

I have to test it, today I'll do that.

alessiodionisi:
Tested today, it works

nutellinoit (approved):
LGTM

alessiodionisi:
Re-requested the reviews due to change on error handling

ralgozino (approved):
lgtm 👍

PR #549, opened 2024-10-21
feat(patches): add compatibility patch for hubble ingress basic auth

ralgozino:
Add compatibility patch for Hubble Ingress when using basic auth as auth provider on KFD versions 1.27.9, 1.28.4, and 1.29.4.
Reference: https://github.com/sighupio/fury-distribution/pull/277


nutellinoit (approved):
LGTM! Ship it! 🚀

PR #554, opened 2024-11-15
feat: enable `kapp` support

kriive:
This PR adds support for `kapp`. It relies on `fury-distribution` to support the schema edits.

If `kapp` is not found in the `kfd.yaml` (or rather, its version is set to "") it won't be downloaded at all.


alessiodionisi (approved):

alessiodionisi:
LGTM

Issue #508, opened 2024-04-24
`--outdir .` flag value makes furyctl not working correctly


alessiodionisi:
When `--outdir` value is `.` furyctl cannot find binaries.

```
$ furyctl apply --debug --outdir . --distro-location ../../fury-distribution --dry-run
DEBU logging to: .furyctl/furyctl.1713949596-73917.log
DEBU Getting Home Directory Path...
INFO Dry run mode enabled, no changes will be applied
INFO Downloading distribution...
DEBU Downloading '/Users/alessiodionisi/Projects/sighupio/fury-distribution' in '/var/folders/qt/7dtl66_55rdc5dpfg3q04jjr0000gn/T/furyctl-4096927723/data'
INFO Validating configuration file...
INFO Downloading dependencies...
DEBU Cleaning vendor folder
INFO Validating dependencies...
INFO Running preflight checks
INFO Checking that the cluster is reachable...
ERRO error while creating cluster: error while executing preflight phase: cluster is unreachable, make sure you have access to the cluster: error getting kubectl version: .furyctl/bin/kubectl/1.28.7/kubectl .furyctl/bin/kubectl/1.28.7/kubectl version -o json: command failed - fork/exec .furyctl/bin/kubectl/1.28.7/kubectl: no such file or directory
out:
err:
```

Temporary fix: `--outdir $PWD`

ralgozino:
May be related #434

ralgozino:
> May be related to [#434](https://github.com/sighupio/furyctl/issues/434)

Kinda related but not they not share the same root cause.

The issue is that furyctl when launches the tools, sometimes changes the working directory from where it launches them. So the relative path set in the outdir does not point to the right place anymore.

This issue is even more evident if you launch furyctl using the `--workdir` flag, then, the relative paths get completely broken.

Passing `$PWD` instead of `.` works because it is an absolute path so the paths are not relative anymore and always work.


We need to convert the outDir to an absolute path early to fix this issue.

Exactly the same issue is present with the `--bin-path`, because when it is not set it is calculated starting from `--outdir`. You can verify this by setting `--outdir` to an absolute path like `$PWD` and `--bin-path` to a relative path, you will get the same error as OP; and viceversa: you can set outdir to a relative path and bin-path to an absolute path and you won't get the error from OP.

PR #557, opened 2024-11-20
feat(cmd): use dot to specify current dir when using --outdir flag

luigidematteis:
This PR solves the issue #508.

Now is possible to use `.` to specify the current directory when using the `--outdir` flag.

In addition to the `--outdir` flag for the `apply` command, the same behavior has been implement also for the other commands that use it, which are: `delete cluster`, `diff`, `connect openvpn` and `dump template`.

## Tests

- [x]  Result of `furyctl apply`:

```
furyctl apply --outdir . --debug
...
Apply script completed.
```

- [x]  Result of  `furyctl dump template`:
```
furyctl dump template --workdir $(PWD) --outdir . --debug 
...       
INFO Distribution manifests generated successfully 
```
- [x] Result of `furyctl diff` command:
```
furyctl diff --workdir $(PWD) -o .
...
INFO No differences found from previous cluster configuration 
```
- [ ] Check `connect openvpn` command:

`Help needed`

- [x] Result of `furyctl delete cluster` command:
```
furyctl delete cluster --workdir $(PWD) --outdir .
...
INFO Kubernetes Fury Distribution deleted successfully
```

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -20,6 +20,7 @@ import (
 	_ "github.com/sighupio/furyctl/internal/apis/kfd/v1alpha2/ekscluster"
 	"github.com/sighupio/furyctl/internal/app"
 	"github.com/sighupio/furyctl/internal/cluster"
+	utils "github.com/sighupio/furyctl/internal/cmdutils"
 	"github.com/sighupio/furyctl/internal/config"
 	"github.com/sighupio/furyctl/internal/git"
 	"github.com/sighupio/furyctl/internal/lockfile"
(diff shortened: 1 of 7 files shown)


g-iannelli (changes requested):
I don't like the change in default behavior for `--output-dir`. Using the user's home directory is safer than using the current directory. No one can guarantee that someone won't run a command like `furyctl dump template -c /dir1/dir2/dir3/furcytl.yaml` and write it to the current directory (which is unknown) instead of using the user's home.

IMHO the user's home should be the default and the check for the `--output-dir == .` should be added in addition.

g-iannelli:
Moreover the fix doen't work in case of `--workdir` is specified 

```
pwd && go run main.go dump template -D -w /tmp -o .
~/Workspace/github.com/luigidematteis/furyctl
DEBU logging to: .furyctl/furyctl.1732113133-33825.log
DEBU Changed working directory to /tmp
DEBU Getting Home Directory Path...
INFO Compatibility patches applied for v1.29.4
INFO Validating configuration file...
INFO Generating distribution manifests...
DEBU config path = /var/folders/k9/3lv0pm253td7mqnyhqfryprw0000gn/T/furyctl-dist-2246877313/config.yaml
DEBU output directory = /private/tmp/distribution
INFO Distribution manifests generated successfully
```

luigidematteis:
Since this PR is stalled, as we don't have an easy way to test the `connect openvpn` command, I close this PR and maybe work on it later if needed

PR #558, opened 2024-11-29
fix(upgrades): delete monitoring resources only when enabled, prepare release 0.30.1

ralgozino:
The upgrade scripts for migrating to ingress v3 tried to delete resources that are not needed anymore even when they may not exist. The flag --ignore-not-found works for native resources, but fails when trying to delete resources that the CRDs are not deployed to the cluster, so the kind is unknown for the API.

Prepare docs for releasing v0.30.1 with this fix.


nutellinoit (approved):
LGTM!

PR #562, opened 2024-12-18
feat: release 0.31.0

kriive:
In this PR we're adding support for fury v1.28.6, v1.29.6, v1.30.1 and v1.31.0.


nutellinoit (approved):
LGTM!

Issue #551, opened 2024-10-24
`validateTools` could panic in case of a non-existing or non-supported tool


kriive:
In the following snippet here, we fail to check for `nil` value, after `Create`ing the tool.
 
https://github.com/sighupio/furyctl/blob/ba7dc1dcfa01ec858def0a6db4b90bb3df70023b/internal/dependencies/tools/validator.go#L125-L126

PR #564, opened 2025-01-27
fix(tools): handle nil tools in validator

stefanoghinelli:
### Summary 💡

This PR introduces a fix to handle cases where a tool returned by the `toolFactory` is `nil` in the `Validator` implementation. 
The change prevents a potential panic in case of a non-existing or non-supported tool.

Closes: [551](https://github.com/sighupio/furyctl/issues/551)

### Description 📝

The potential issue occurred because `Validator.validateTools` did not check if `toolFactory.Create` returned `nil`. 

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested the change with tools returning `nil` to verify the error handling
- [x] Ran locally integration tests

### Future work 🔧

None

### DIFF
--- a/internal/dependencies/tools/validator.go
+++ b/internal/dependencies/tools/validator.go
@@ -17,7 +17,10 @@ import (
 	execx "github.com/sighupio/furyctl/internal/x/exec"
 )
 
-var ErrWrongToolVersion = errors.New("wrong tool version")
+var (
+	ErrWrongToolVersion = errors.New("wrong tool version")
+	ErrToolNotFound     = errors.New("tool not found in the toolFactory, possible fury-distribution version mismatch")
+)
 
 func NewValidator(executor execx.Executor, binPath, furyctlPath string, autoConnect bool) *Validator {
 	return &Validator{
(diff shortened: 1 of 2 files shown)


kriive (dismissed):
LGTM!

internal/dependencies/tools/validator.go
```diff
@@ -127,6 +131,16 @@ func (tv *Validator) validateTools(i any, kfdManifest config.KFD) ([]string, []e
 		}
 
 		tool := tv.toolFactory.Create(itool.Name(toolName), toolCfg.Version)
+
+		if tool == nil {
+			errs = append(
+				errs,
+				fmt.Errorf("%w: %s", ErrToolNotFound, toolName),
```
ralgozino, line 138: I think that the order should be inverted for better readability and the version that we are checking included in the error message too

```suggestion
				fmt.Errorf("%s version %s: %w", toolName, ErrToolNotFound),
```

ralgozino (changes requested):

internal/dependencies/tools/validator.go
stefanoghinelli: Let me check :)

ralgozino (approved):
lgtm

Issue #565, opened 2025-02-05
Bug: apply gives a warning about unsupported version with OnPremises provider and version 1.30.1


speziato:
**Describe the bug**
`furyctl apply` gives a warning when using `distributionVersion: v1.30.1` with the `OnPremises` provider, then continues without issues:

```bash
$ furyctl create config --kind OnPremises --version v1.30.1
INFO Downloading distribution...                  
INFO Configuration file created successfully at: /Users/rcannella/projects/local/test/furyctl.yaml
$ furyctl apply                                            
INFO Downloading distribution...                  
WARN The specified KFD version v1.30.1 is not supported by furyctl, please upgrade furyctl to the latest version or use a supported version
INFO Validating configuration file...             
INFO Downloading dependencies...
...
```

**To Reproduce**
1. Run `furyctl generate config --kind OnPremises --version v1.30.1`
2. Run `furyctl apply`
3. See the warning

**Expected behavior**
`furyctl` applies the configuration without giving warnings when using supported versions.

**Screenshots**

![Image](https://github.com/user-attachments/assets/97a29a62-6209-404d-9c8a-275dae80ae07)

**Desktop (please complete the following information):**
 - OS: macOS Sequoia 15.3
 - Board: MacBook Pro 2019
 - Architecture: amd64

**Kubernetes (please complete the following information):**
 - Kubernetes version: 1.30.X

PR #566, opened 2025-02-05
fix(compatibility): use the right max130Version for OnPremises

speziato:
### Summary 💡

This PR solves a small bug where `furyctl apply` commands with the OnPremises provider and version 1.30.1  give a warning about v1.30.1 being unsupported, when it's not the case.

This bug appears only with the OnPremises provider, while EKSCluster and KFDDistribution do not show this issue.

Closes: #565 

### Description 📝

The change is a rather small one, adjusting the `max130Version` to the correct one in the code block that does the compatibility checks for the OnPremises provider. I also added relative unit tests to verify this behaviour.

### Breaking Changes 💔

None.
### Tests performed 🧪

- [x] Tested the change with a real KFD OnPremises version 1.30.1 configuration file
- [x] Tested with dedicated unit tests for all providers
### Future work 🔧

N/A

### DIFF
--- a/internal/distribution/compatibility.go
+++ b/internal/distribution/compatibility.go
@@ -306,7 +306,7 @@ func (c *OnPremisesCheck) IsCompatible() bool {
 		return false
 	}
 
-	max130Version, err := semver.NewVersion("v1.30.0")
+	max130Version, err := semver.NewVersion("v1.30.1")
 	if err != nil {
 		return false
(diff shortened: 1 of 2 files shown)


kriive (approved):
LGTM! Nice catch!

ralgozino (approved):
SBAM!

Issue #210, opened 2023-02-07
Add command to furyctl to print all KFD available versions


g-iannelli:
Add command to furyctl to print all KFD available versions.

E.g. `furyctl info distro versions`

Filo01:
@nutellinoit 

Questions about the implementation:
- do we want to use `furyctl info distro versions` or something like else?
- do we want the versions shown to be about what's deployed on the cluster, or what's inside the `furyctl.yaml` ?
  - in case we want to show what's deployed on the cluster, do we want to also show a notification message if there are unapplied modifications to the `furyctl.yaml`? (eg. _Warning changes to the furyctl.yaml detected, use `furyctl diff` to show the diff_ )
- do we want to **eventually** expand the `furyctl info` command with the other phases? (eg. `furyctl info kubernetes versions`, `furyctl info plugins versions`...)
- do we want to **eventually** add other informations other than `version` in the command?
- where do we lookup the applied configuration? from the secret `furyctl-config.config` or `furyctl-config.rendered`?
---
Proposed example output of command:
```bash
> furyctl info distro versions
INFO KFD distribution: v1.30.0
INFO KFD distribution type: OnPremises
INFO networking.calico: vX.X.X
INFO logging.prometheus: vX.X.X
INFO tracing.tempo: vX.X.X
INFO policy.gatekeeper: vX.X.X
INFO dr.velero: vX.X.X
INFO auth.pomerium: vX.X.X
```

Filo01:
ok, so we discussed it via chat with @nutellinoit 
Let's disregard my previous comment.

This command should not show the modules version, only the available distro versions, without the need of an active kubernetes cluster. Basically it should fetch the fury-distribution tags on github except RCs.

Given this information I would propose a command like `furyctl search distro` similar to helm's `helm repo search [NAME]` so it's easier to understand that it's not an action on a cluster, but a search from an external source.

## Proposed output


```bash
> furyctl search distro

AVAILABLE KUBERNETES FURY DISTRIBUTION VERSIONS  
------------------------------------------------  
VERSION     RELEASE DATE   STATUS      NOTES                    
v1.31.2     2024-12-15     stable      Latest stable release    
v1.31.1     2024-11-15     stable      Upgrade recommended      
v1.31.0     2024-10-15     stable      Upgrade recommended      
v1.30.2     2024-09-15     stable      Supported version        
v1.30.1     2024-08-15     stable      Upgrade recommended      
[...]

PR #567, opened 2025-02-06
feat: add new `furyctl get distro-versions` command

Filo01:
### Summary 💡

Created new subcommand `furyctl get distro-versions` to show available supported versions for the fury-distribution. The command only shows supported versions and if the various distribution kinds are compatible with the furyctl cli

Closes: #210 

### Description 📝

- Added new command `furyctl get distro-versions`
- Added a way to query directly the public github api to retrieve sighupio/kubernetes-fury tags and commit info
- Added logic to check 

### Breaking Changes 💔

None

### Tests performed 🧪

- Unit tests (`cmd/get/distro-versions_test.go`, `internal/app/distro_versions_test.go`)
- run `furyctl get distro-versions`
Output:
```
INFO AVAILABLE KUBERNETES FURY DISTRIBUTION VERSIONS
-----------------------------------------------
VERSION	RELEASE DATE	EKS	KFD	ON PREMISE
v1.29.0	2024-04-24	X	X	X
v1.29.1	0001-01-01	X	X	X
v1.29.2	2024-08-09	X	X	X
v1.29.3	2024-08-27	X	X	X
v1.29.4	2024-09-20	X	X	X
v1.29.5	0001-01-01	X	X	X
v1.29.6	0001-01-01	X	X	X
v1.30.0	0001-01-01	X	X	X
v1.30.1	0001-01-01	X	X	X
v1.31.0	0001-01-01	X	X	X 
```

### Future work 🔧

- The way it checks for compatibility for various distribution's kinds the should be more customisable to be able to be tested properly
- The logic to get the last supported version is very basic (latest version minor and the 2 previous, for example 1.31.0 minus 2 minor versions = 1.29.0) which would probably break in case of a major version bump

### DIFF
--- a/cmd/get/distro-versions.go
+++ b/cmd/get/distro-versions.go
@@ -1,0 +1,80 @@
+// Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package get
+
+import (
+	"fmt"
+
+	"github.com/sighupio/furyctl/internal/analytics"
+	"github.com/sighupio/furyctl/internal/app"
+	"github.com/sighupio/furyctl/internal/git"
+	cobrax "github.com/sighupio/furyctl/internal/x/cobra"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const DateFmt = "2006-01-02"
+
+func NewDistroVersionCmd() *cobra.Command {
+	var cmdEvent analytics.Event
+
+	distroVersionCmd := &cobra.Command{
+		Use:   "distro-versions",
+		Short: "Get the supported distro versions and shows compatibilities with the current furyctl version used to invoke this command with the different distribution's kind.",
+		PreRun: func(cmd *cobra.Command, _ []string) {
+			cmdEvent = analytics.NewCommandEvent(cobrax.GetFullname(cmd))
+
+			if err := viper.BindPFlags(cmd.Flags()); err != nil {
+				logrus.Fatalf("error while binding flags: %v", err)
+			}
+		},
+		RunE: func(_ *cobra.Command, _ []string) error {
+			ctn := app.GetContainerInstance()
+
+			tracker := ctn.Tracker()
+			tracker.Flush()
+			releases, err := app.GetSupportedDistroVersions(git.NewGitHubClient())
+			if err != nil {
+				cmdEvent.AddErrorMessage(err)
+				tracker.Track(cmdEvent)
+				return fmt.Errorf("error getting supported distro versions: %w", err)
+			}
+			logrus.Info(formatDistroVersions(releases))
+
+			cmdEvent.AddSuccessMessage("upgrade paths successfully retrieved")
+			tracker.Track(cmdEvent)
+
+			return nil
+		},
+	}
+
+	return distroVersionCmd
+}
+
+func formatDistroVersions(releases []app.DistroRelease) string {
+	fmtDistroVersions := ""
+	fmtDistroVersions += "AVAILABLE KUBERNETES FURY DISTRIBUTION VERSIONS\n"
+	fmtDistroVersions += "-----------------------------------------------\n"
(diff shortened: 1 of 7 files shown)


internal/app/distro_versions.go
ralgozino, line 1: I'm curious, why did you base all the logic on the Tags API instead of using the Releases API?

https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#list-releases

internal/app/distro_versions.go
```diff
@@ -0,0 +1,206 @@
...
+		commitDate, err = time.Parse(time.RFC3339, commit.Author.Date)
+		if err != nil {
+			commitDate = time.Time{}
+		}
+	} else if commit.Tagger != nil {
+		commitDate, err = time.Parse(time.RFC3339, commit.Tagger.Date)
+		if err != nil {
+			commitDate = time.Time{}
+		}
+	}
```
ralgozino, line 149: You are using the zero value (`0001-01-01 00:00:00 +0000 UTC`) here as a fallback but I don't see logic in the presentation to show something more relevant to the user instead of the zero time (may be I missed it).

ralgozino (changes requested):

internal/app/distro_versions.go
Filo01: Because I forgot it existed, I'll fix it

cmd/get/distro-versions.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	return distroVersionCmd
+}
+
+func FormatDistroVersions(releases []app.DistroRelease) string {
+	fmtDistroVersions := ""
+	fmtDistroVersions += "AVAILABLE KUBERNETES FURY DISTRIBUTION VERSIONS\n"
+	fmtDistroVersions += "-----------------------------------------------\n"
+	fmtDistroVersions += "VERSION\tRELEASE DATE\tEKS\tKFD\tON PREMISE\n"
```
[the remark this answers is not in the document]
ralgozino: Yes, see the internal/distribution/compatibility.go file

cmd/get/distro-versions.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+const DateFmt = "2006-01-02"
+
+func NewDistroVersionCmd() *cobra.Command {
+	var cmdEvent analytics.Event
+
+	distroVersionCmd := &cobra.Command{
+		Use:   "distro-versions",
+		Short: "Get the supported distro versions and shows compatibilities with the current furyctl version used to invoke this command with the different distribution's kind.",
```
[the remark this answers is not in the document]
Filo01: ok, I've added the flag `--kind`

```
furyctl get supported-versions --kind EKSCluster

INFO 
------------------------------------------------------------------------------------
VERSION		RELEASE DATE		EKSCluster	
------------------------------------------------------------------------------------
v1.31.0		2024-12-24		Yes
v1.30.1		2025-01-16		Yes
v1.30.0		2024-11-27		Yes
v1.29.6		2025-01-16		Yes
v1.29.5		2024-11-29		Yes
v1.29.4		2024-09-20		Yes
v1.29.3		2024-08-27		Yes
v1.29.2		2024-08-09		Yes
v1.29.1		2024-05-09		Yes
v1.29.0		2024-04-24		Yes 
```

cmd/get/supported-versions.go
```diff
@@ -104,5 +126,9 @@ func FormatSupportedVersions(releases []distribution.KFDRelease, kinds []string)
 		fmtSupportedVersions += "\n"
 	}
 
+	if showUnsupportedFuryctlMsg {
+		fmtSupportedVersions += "* this usually indicates you are not using the latest version of furyctl, try updating or checking the online documentation.\n"
```
ralgozino, line 135: let's add an empty before the line, otherwise it gets printed right after the table
```suggestion
		fmtSupportedVersions += "\n* this usually indicates you are not using the latest version of furyctl, try updating or checking the online documentation.\n"
```

ralgozino (changes requested):
looking good! some minor tweaks to formatting

ralgozino (approved):
LGTM, well done! 🚀

PR #586, opened 2025-04-10
fix(upgrade-paths): delete old loki resources

ralgozino:
### Summary 💡

Delete resources that are not needed anymore after upgading to logging v5, that updates and deprecates some Loki components.

Relates: https://github.com/sighupio/module-logging/pull/185

### Description 📝

In logging v5 some components of Loki have changed kind or name. We need to do some clean up after the upgrade when we are coming from versions that were not using `kapp`.

In versions that were previously using `kapp` this was not needed.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested the 1.30.1 -> 1.30.2 upgrade path in an on-premises cluster with Loki enabled. Clean up was successful.

### Future work 🔧

None


kriive (approved):
LGTM!

lukeskywlkr (approved):
LGTM 🚀

PR #580, opened 2025-03-24
Feat: rebranding

nutellinoit:
### Summary 💡

This PR updates all the occurrences for Kubernetes Fury Distribution and replace them with SIGHUP Distribution.

A little bit of code has been touched, but only strings that are printed to the user.

PR #587, opened 2025-04-11
chore: more rebranding

ralgozino:
### Summary 💡

- Simplify readme, dropping information that we already have in the docs, so we don't have to maintain both versions.
- Fix some missing rebranding in README
- Fix some missing rebranding in the logs printed by furyctl
- Improve some of the logs that were leaking implementation details

Relates: #580 

### Description 📝

See above

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] build and tested a furyctl apply and upgrade.

### Future work 🔧

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,28 +15,30 @@
 </h1>
 <!-- markdownlint-eable MD033 -->
 
-<!-- <KFD-DOCS> -->
-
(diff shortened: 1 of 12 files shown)


kriive (approved):
LGTM!

PR #593, opened 2025-04-24
feat: add dump cli-reference command

ralgozino:
### Summary 💡

Add a new `dump cli-reference` command that exports the full CLI reference in Markdown to the filesystem.

The command accepts one argument with the path of a folder in the working dir to where save the files.

This command is useful to keep our docs site updated easily.

### Description 📝

- Added logic for the new `dump cli-reference` command.
- Improved some help messages of the other commands so they make sense in the docs.
- Fixed some missing rebranding.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested the help message
- [x] Tested that the command outputs the expected files in the expected folder.

### Future work 🔧

We had to take some functions from cobra's code base and adapt them to our needs. Doing this I realised that that code can be improved even further and maybe we can move some logic that we have in the command to the helper functions.

Also, it could be better to use templates instead of a gazillion writes to the buffer (with the corresponding gazillion check for write errors to make the linter happy).


cmd/dump/cli-reference.go
```diff
@@ -0,0 +1,194 @@
...
+				}
+
+				return name
+			}
+
+			cmd.Root().DisableAutoGenTag = true
+			mainFile.WriteString(`---
+title: furyctl CLI Reference
+sidebar_position: 4
+---
```
speziato, line 131: I think this is an implementation detail of the documentation site, I'm not sure it should be placed here

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,195 @@
...
+		return err
+	}
+	if hasSeeAlso(cmd) {
+		buf.WriteString("## See Also\n\n")
+		if cmd.HasParent() {
+			parent := cmd.Parent()
+			pname := parent.CommandPath()
+			link := pname + markdownExtension
+			link = strings.ReplaceAll(link, " ", "_")
+			buf.WriteString(fmt.Sprintf("* [%s](%s)\t - %s\n", pname, linkHandler(link), parent.Short))
```
speziato, line 111:
```suggestion
			buf.WriteString(fmt.Sprintf("* [%s](%s) - %s\n", pname, linkHandler(link), parent.Short))
```

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,195 @@
...
+		sort.Sort(byName(children))
+
+		for _, child := range children {
+			if !child.IsAvailableCommand() || child.IsAdditionalHelpTopicCommand() {
+				continue
+			}
+			cname := name + " " + child.Name()
+			link := cname + markdownExtension
+			link = strings.ReplaceAll(link, " ", "_")
+			buf.WriteString(fmt.Sprintf("* [%s](%s)\t - %s\n", cname, linkHandler(link), child.Short))
```
speziato, line 129:
```suggestion
			buf.WriteString(fmt.Sprintf("* [%s](%s) - %s\n", cname, linkHandler(link), child.Short))
```

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,195 @@
...
+		}
+		buf.WriteString("\n")
```
speziato, line 131: this line can be deleted, as every generated text already has a newline and this one is just an additional one that actually makes linters go mad
```suggestion

```

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,195 @@
...
+	cmd.InitDefaultHelpFlag()
+
+	buf := new(bytes.Buffer)
+	name := cmd.CommandPath()
+
+	buf.WriteString("# " + name + "\n\n")
+	buf.WriteString(cmd.Short + "\n\n")
+	if len(cmd.Long) > 0 {
+		buf.WriteString("## Synopsis\n\n")
+		buf.WriteString(cmd.Long + "\n\n")
```
speziato, line 88:
```suggestion
		buf.WriteString(strings.ReplaceAll(cmd.Long, "\t", "") + "\n\n")
```

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if cmd.Runnable() {
+		buf.WriteString("## Usage\n\n")
+		buf.WriteString(fmt.Sprintf("```bash\n%s\n```\n\n", cmd.UseLine()))
```
speziato, line 93:
```suggestion
		buf.WriteString(fmt.Sprintf("```bash\n%s\n```\n\n", strings.ReplaceAll(cmd.UseLine(), "\t", "")))
```

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	if len(cmd.Example) > 0 {
+		buf.WriteString("## Examples\n\n")
+		buf.WriteString(fmt.Sprintf("```bash\n%s\n```\n\n", cmd.Example))
```
speziato, line 98:
```suggestion
		buf.WriteString(fmt.Sprintf("```bash\n%s\n```\n\n", strings.ReplaceAll(cmd.Example, "\t", "")))
```

speziato (changes requested):
Added a couple of comments in the code. Also, the same approach for the examples in `cmd/get/upgrade-paths.go` should be used in `cmd/get/supported-versions.go`

cmd/dump/cli-reference.go
ralgozino: This command is for our use, I don't think an enduser will actually use it. Anyways, it does not affect the rendering of the markdown, a user can modify it if needed.

ralgozino: It turns out we don't need the front matter any ways :D

cmd/dump/cli-reference.go
```diff
@@ -0,0 +1,186 @@
...
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read file %s: %w", path, err)
+	}
+
+	reUnix := regexp.MustCompile(`(.+)\$ (.+)`)
+	rePS := regexp.MustCompile(`(.+)PS> (.+)`)
+	escapedContent := string(content)
+	escapedContent = reUnix.ReplaceAllString(escapedContent, "${1}```shell\n${1}$$ ${2}\n${1}```")
+	escapedContent = rePS.ReplaceAllString(escapedContent, "${1}```powershell\n${1}PS> ${2}\n${1}```")
```
speziato, line 174: this is a fix for a linting error inside the generated MD where codeblocks are not properly sorrounded by blank lines (e.g.: `furyctl completion`), but it makes sense to maintain that styling for the CLI help messages:

```suggestion
	escapedContent = rePS.ReplaceAllString(escapedContent, "${1}```powershell\n${1}PS> ${2}\n${1}```")
        reFencedNewLineFix := regexp.MustCompile(":\\n(.+)```")
        escapedContent = reFencedNewLineFix.ReplaceAllString(escapedContent, ":\n\n${1}```")
```

speziato (changes requested):
Just a small fix to avoid making markdownlint angry

speziato (dismissed):
LGTM! 🔥

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,247 @@
...
+	return nil
+}
+
+// GenMarkdown creates markdown output.
+func GenMarkdown(cmd *cobra.Command, w io.Writer) error {
+	return GenMarkdownCustom(cmd, w, func(s string) string { return s })
+}
+
+// GenMarkdownCustom creates custom markdown output.
+func GenMarkdownCustom(cmd *cobra.Command, w io.Writer, linkHandler func(string) string) error {
```
kriive, line 92: I would leave this as unexported, and only interact with the "public API" through the various `GenMarkdown` specialized functions. What do you think?

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,247 @@
...
+func GenMarkdownTree(cmd *cobra.Command, dir string) error {
+	identity := func(s string) string { return s }
+	emptyStr := func(_ string) string { return "" }
+
+	return GenMarkdownTreeCustom(cmd, dir, emptyStr, identity)
+}
+
+// GenMarkdownTreeCustom is the same as GenMarkdownTree, but
+// with custom filePrepender and linkHandler.
+func GenMarkdownTreeCustom(cmd *cobra.Command, dir string, filePrepender, linkHandler func(string) string) error {
```
kriive, line 195: I would leave this unexported and call it `genMarkdownTree`, to highlight that this is the underlying implementation for `GenMarkdownTree`.

cmd/dump/cmd-to-md.go
```diff
@@ -0,0 +1,247 @@
...
+		if _, err := buf.WriteString("```\n\n"); err != nil {
+			return fmt.Errorf("error while writing to buffer: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// GenMarkdown creates markdown output.
+func GenMarkdown(cmd *cobra.Command, w io.Writer) error {
```
kriive, line 87: Do we really need to pass an `io.Writer` here? Shouldn't we just output a string and then leave to the caller the decision of what to do with it?

kriive (commented):
This looks good! I only left some comments pertaining the code organization. Let me know you thoughts on these.

cmd/dump/cmd-to-md.go
ralgozino: Removed function altogether. We were not using it

cmd/dump/cmd-to-md.go
ralgozino: I'll leave the name as-is, but not exported. To have a reference do the original methods until we change the logic to something simpler.

kriive (approved):
LGTM!

PR #599, opened 2025-04-30
feat: print lock file path

ralgozino:
### Summary 💡

Print also the path to the lock file in the messages shown to the user.

So, if your furyctl execution is blocked by the lock file by mistake (a dead process that did not clean up the lock, for example) the user knows which file to delete.

### Description 📝

See summary.

Example output:

```
❯ furyctl apply
INFO Compatibility patches applied for v1.31.1
ERRO error while verifying lock file /var/folders/lb/zscvwt_s6fx6qdrfv4g51dsc0000gn/T/furyctl-multipass: lock file exists. This usually means that there is another instance of furyctl running
```

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested to run 2 furyctl instances in parallel and verified that the path is added to the error message.

### Future work 🔧

It would be nice if the lock file had the PID inside.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -184,7 +184,7 @@ func NewApplyCmd() *cobra.Command {
 					logrus.Debugf("Removing lock file %s", lockFileHandler.Path)
 
 					if err := lockFileHandler.Remove(); err != nil {
-						logrus.Errorf("error while removing lock file: %v", err)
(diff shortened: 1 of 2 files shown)


stefanoghinelli (commented):
Very good! I will wait for the CI then approve :)

PR #601, opened 2025-05-05
feat: add tab-completion for known flags

ralgozino:
### Summary 💡

- Add tab completion for known values of the flags, like the options for the phase flag.
- Improve version checking for the create config command.
- A little refactoring to not repeat ourselves in where are these values defined.

### Description 📝

Improve UX by allowing to tab-complete valid options for flags that take specific values like --phase, --start-from, --kind an similar.

I also took the chance to do a little refactoring so we start to have some unique source of truth for some of this things.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested that all the modified flags tab-autocomplete the right values
- [x] Tested that an on-prem cluster creation still works.
- [x] Tested that an on-prem cluster upgrade still works.

### Future work 🔧

In general the code structure is not clear to me, I believe that we can improve and better organize it, having some common centralized definitons instead of repeating our selves all over the place.

For example, the list of valid Kinds should be available in the APIs module instead of having it inside the compatibility.go file. Speaking of the API module, we seem to have 2? One in the distribution's repository and one here in furyctl's, I don't get why.

### DIFF
--- a/internal/cluster/phase.go
+++ b/internal/cluster/phase.go
@@ -50,55 +50,67 @@ var (
 		"unsupported operation phase, options are: pre-infrastructure, infrastructure, post-infrastructure, " +
 			"pre-kubernetes, kubernetes, post-kubernetes, pre-distribution, distribution, post-distribution, plugins",
 	)
-	ErrChangesToOtherPhases = errors.New("changes to other phases detected")
+	ErrChangesToOtherPhases = errors.New("changes to other phases detected. When using the --phase flag, changes " +
+		"only to the section corresponding to the selected phase are accepted. ",
+	)
 )
 
 func CheckPhase(phase string) error {
-	switch phase {
-	case OperationPhasePreFlight,
+	phases := slices.Concat(
+		GetMainPhases(),
+		[]string{
+			OperationPhasePreFlight,
+			OperationPhaseAll,
+		})
+	if slices.Contains(phases, phase) {
+		return nil
+	}
+
+	return ErrUnsupportedPhase
+}
+
+// GetMainPhases returns all the main phases that can be used in the operation phase.
+func GetMainPhases() []string {
+	return []string{
 		OperationPhaseInfrastructure,
 		OperationPhaseKubernetes,
 		OperationPhaseDistribution,
 		OperationPhasePlugins,
-		OperationPhaseAll:
-		return nil
-
-	default:
-		return ErrUnsupportedPhase
 	}
 }
 
-func ValidateOperationPhase(phase string) error {
-	err := CheckPhase(phase)
-	if err == nil {
-		return nil
-	}
-
-	switch phase {
-	case OperationSubPhasePreInfrastructure,
+// GetOperationPhases returns all the sub-phases that can be used in the operation phase.
+func GetOperationPhases() []string {
+	return []string{
+		OperationSubPhasePreInfrastructure,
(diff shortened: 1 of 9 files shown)


internal/distribution/compatibility.go
```diff
@@ -26,6 +27,28 @@ type VersionRange struct {
...
+	// but AFAIK we don't have a way to do that yet.
+	return []string{
+		EKSClusterKind,
+		KFDDistributionKind,
+		OnPremisesKind,
+	}
+}
+
+// ValidateConfigKind checks if the given kind is supported and returns the normalised value for the Kind and an error.
+func ValidateConfigKind(kind string) (string, error) {
```
kriive, line 42: Maybe move the validation functions inside a new file named `internal/distribution/kind.go`? Or something like that, to separate it from the `compatibility.go`. I am super open to hear your opinions on this.

kriive (changes requested):
Few nitpicks a some questions on the general code organization, other than that LGTM! 🚀

internal/cluster/phase.go
```diff
@@ -50,55 +50,67 @@ var (
...
+		})
+	if slices.Contains(phases, phase) {
+		return nil
+	}
+
+	return ErrUnsupportedPhase
+}
+
+// GetMainPhases returns all the main phases that can be used in the operation phase.
+func GetMainPhases() []string {
```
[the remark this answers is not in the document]
ralgozino: sounds good, I'll rename this and the other funcs

internal/distribution/compatibility.go
ralgozino: I did NOT want to do a major refactor with this feature, this PR should have been even simpler. But I took the chance to centralised at least some definitions.

I'm fine with moving them to a separated file. I evaluated several places but all felt a little odd or caused circular import issues that would have meant to do more refactor to fix.

Ideally, I think that this (the available Kinds and related funcs) should be somewhere in the distribution's API module instead or the equivalent to it in this repo (that's where the configuration kinds come from), but again doing that would have ended up in a scope creep of this PR.

Overall the module structure of furyctl (and the distribution's go modules) is not clear to me.

internal/cluster/phase.go
```diff
@@ -50,55 +50,67 @@ var (
...
+	return ErrUnsupportedOperationPhase
 }
 
 func ValidateMainPhases(phase string) error {
-	switch phase {
-	case OperationPhaseInfrastructure,
-		OperationPhaseKubernetes,
-		OperationPhaseDistribution,
-		OperationPhasePlugins:
+	if slices.Contains(GetMainPhases(), phase) {
```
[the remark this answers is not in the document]
ralgozino: I thought about it but I did not do that on purpose, I just simplified the logic to use the common definitions but I did not want to change the functionality to avoid causing unwanted change in behaviour that is out of the scope of this PR.

I agree that it would be better to have this case-insensitive, but I'd do that in a different PR.

(Phases are all lower-case btw, Kinds have mixed case like "OnPremises", so the "risk" of not using the write case is lowish)

internal/distribution/compatibility.go
kriive: Yeah, I agree that we need to clarify the whole architecture a bit more and maybe have some central design documents available as a reference.

kriive (dismissed):
LGTM 🚀

Issue #434, opened 2023-10-06
furyctl-ng:  dump command --outdir param check


davide-rossi-ce:
verified on furyctl 0.26.1

The --outdir behaviour of the dump command( that writes directly to the directory passed as --outdir param ) is different from other commands ( that instead work on the .furyctl child dir of --outdir ).

Also, the dump command seems to **wipe the --outdir directory** before doing any other action. 

This can lead to mistakenly wipe the project dir. 
Maybe a check could be done to verify if the dump outdir is empty, 
or maybe write to another child dir ( es. outdir/.furyctl/dump ).

ralgozino:
I did some tests, what is happening with v0.31.1 is:

- running `furyctl dump template` will create the `.furyctl` folder in the `$HOME` folder and the `distribution` folder with the rendered temlpates in the current directory.
- running `furyctl dump temlpate --outidir <outdir>` will put both the `.furyctl` and the `distribution` folder inside the `<outdir>` folder.
- running `furyctl dump template --workdir <workdir>` will create the `.furyctl` folder in the `$HOME` folder and the `distribution` folder with the rendered temlpates in the `<workdir>` folder, as expected.
- running `furyctl dump template --workdir <workdir> --outdir <outdir>` will put both the `.furyctl` and the `distribution` folders inside the `<outdir>` folder.

> [!NOTE]
> Running the command again does not overwrite the `distribution` folder though. It stops with the following message:
> ```
> ERRO output directory is not empty, set --no-overwrite=false to overwrite it
> ```
> This message could be improved, though.

This is not correct behaviour IMO.

PR #603, opened 2025-05-05
fix(dump template): use workdir instead of outputDir

ralgozino:
### Summary 💡

Use the workdir instead of the outputDir for target forlder where to dump the rendered template.

Fixes: #434 

### Description 📝

Previously the `dump template` command was using _sometimes_ the `outputDir` as the target where to dump the `distribution` folder containing the rendered templates.

With this change, the `dump template` command will create the `distribution` folder in the working directory and dump the rendered templates there always.

Note that the workdir can be set using the `--workdir` flag, like in the other commands.

### Breaking Changes 💔

If you relied on the command dumping to the output dir you may need to adjust the flags you are using.

### Tests performed 🧪

- [x] Tested the the command without flags works as expected: dumps to the workdir and the cache dir is created under the outdir.
- [x] Tested that specifying the `workdir` makes the command create the `distribution` folder in the path specified instead of where the command was launched from.
- [x] Tested the the no-overwrite flag is still working as expected.
- [x] Tested changing outdir to a custom folder and checked that the dump is not done inside it.
- [x] Tested setting both outdir and workdir and verified that the outdir is used only for internal files and that the `distribution` folder is dumped into the workdir.

### Future work 🔧

I think that naming the outdir outdir is confusing, the outdir is a "datadir" and not an output dir. This directory is used only for internal files.

### DIFF
--- a/cmd/dump/template.go
+++ b/cmd/dump/template.go
@@ -51,7 +51,7 @@ func NewTemplateCmd() *cobra.Command {
 		Use:   "template",
 		Short: "Renders the distribution's code from template files parametrized with the configuration file",
 		Long: `Generates a folder with the parametrized version of the Terraform and Kustomization code for deploying the SIGHUP Distribution into a cluster.
-The command will dump into a 'distribution' folder in the working diretory all the rendered files using the parameters set in the configuration file.`,
+The command will dump into a 'distribution' folder in the working directory all the rendered files using the parameters set in the configuration file.`,
 		SilenceUsage:  true,
 		SilenceErrors: true,
 		PreRun: func(cmd *cobra.Command, _ []string) {
--- a/cmd/dump/template.go
+++ b/cmd/dump/template.go
@@ -34,6 +34,7 @@ type TemplateCmdFlags struct {
 	SkipValidation        bool
 	GitProtocol           git.Protocol
 	Outdir                string
+	Workdir               string
 	FuryctlPath           string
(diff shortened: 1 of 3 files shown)


cmd/dump/template.go
```diff
@@ -50,8 +50,8 @@ func NewTemplateCmd() *cobra.Command {
 	templateCmd := &cobra.Command{
 		Use:   "template",
 		Short: "Renders the distribution's code from template files parametrized with the configuration file",
-		Long: `Generates a folder with the Terraform and Kustomization code for deploying the SIGHUP Distribution into a cluster.
-The generated folder will be created starting from a provided templates folder and the parameters set in a configuration file that is merged with default values.`,
+		Long: `Generates a folder with the parametrized version of the Terraform and Kustomization code for deploying the SIGHUP Distribution into a cluster.
+The command will dump into a 'distribution' folder in the working diretory all the rendered files using the parameters set in the configuration file.`,
```
Filo01, line 54:
```suggestion
The command will dump into a 'distribution' folder in the working directory all the rendered files using the parameters set in the configuration file.`,
```

Filo01 (commented):
added a small comment for a typo, the rest LGTM

cmd/dump/template.go
ralgozino: nice catch, thanks!

Filo01 (approved):

PR #604, opened 2025-05-06
feat(lockfile): save PID of running process

ralgozino:
### Summary 💡

- Save the PID of the current furyctl process in the lockfile.
- Print it out the saved PID if the lock file is still present.

Relates: improves on #599

### Description 📝

Show the user the last PID of furyctl when the lockfile is still present, so they can check if the process is still running or the lockfile can safely be removed.

The new error message looks like this:

```
ERRO error while verifying lock file /var/folders/lb/zscvwt_s6fx6qdrfv4g51dsc0000gn/T/furyctl-multipass: lock file exists. Last execution finished abnormally or there may be another instance of furyctl running with PID 70457
```

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Checked that the lockfile mechanism is still working as expected
- [x] Checked that the PID is properly written and valid.

### Future work 🔧

None

### DIFF
--- a/internal/lockfile/lockfile.go
+++ b/internal/lockfile/lockfile.go
@@ -9,10 +9,11 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 )
 
 var ErrLockFileExists = errors.New(
-	"lock file exists. This usually means that there is another instance of furyctl running",
+	"lock file exists. Last execution finished abnormally or there may be another instance of furyctl running with PID",
 )
 
 type LockFile struct {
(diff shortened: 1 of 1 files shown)


internal/lockfile/lockfile.go
```diff
@@ -28,21 +29,21 @@ func NewLockFile(clusterName string) *LockFile {
...
+		return fmt.Errorf("error while checking lock file: \"%w\"", err)
 	}
 
 	return nil
 }
 
 func (l *LockFile) Create() error {
-	_, err := os.Create(l.Path)
-	if err != nil {
+	const perms = os.FileMode(0o666)
```
Al-Pragliola, line 45: :wave: @ralgozino there's a useful set of consts that you can use from https://github.com/sighupio/furyctl/blob/main/internal/x/io/fs.go#L24

ralgozino: Ciao @Al-Pragliola!

Nice! I'll switch to that. 

Thank you! 🫂

PR #611, opened 2025-05-16
fix: fromnodes should check nodes even if they did not change

Filo01:
### Summary 💡

There was a problem with the `fromNodes` reducer in the immutability rules. If the referenced node did not change, the reducer did not have a way to check the current value.

Closes: https://github.com/sighupio/distribution/issues/398

### Description 📝

With this PR the `fromNodes` reducer can work even if the referenced node did not change.
for example with this rule:
```yaml
- path: .spec.distribution.modules.logging.loki.tsdbStartDate
    immutable: true
    description: "changes to Loki's TSDB start date have been detected. This parameter cannot be changed once set."  
    safe:
      - fromNodes:
        - path: ".spec.distribution.modules.logging.type"
          from: "none"
        - path: ".spec.distribution.modules.logging.type"
          to: "none"
        - path: ".spec.distribution.modules.logging.type"
          from: "opensearch"
        - path: ".spec.distribution.modules.logging.type"
          to: "opensearch"
```
changing this:
```
logging:
        type: none
        loki:
          tsdbStartDate: "2025-01-02"
```
to this:
```
logging:
        type: none
        loki:
          tsdbStartDate: "2025-01-03"
```
should still work

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] in a kfd cluster changing `logging.loki.tsdbStartDate: "2025-01-02"` to `logging.loki.tsdbStartDate: "2025-01-03"` while maintaing `logging.type: "none" in both cases -> does not get blocked
- [x] regression test in a kfd cluster changing `logging.type: none` to `logging.type: loki` -> does not get blocked
- [x] regression test in a kfd cluster changing `logging.type: loki` to `logging.type: none` -> does not get blocked
- [x] regression test in a kfd cluster changing `logging.loki.tsdbStartDate: "2025-01-02"` to `logging.loki.tsdbStartDate: "2025-01-03"` while "logging.type: "loki"` -> gets blocked

### Future work 🔧

None

### DIFF
--- a/internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
+++ b/internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
@@ -305,7 +305,7 @@ func (p *PreFlight) CreateDiffChecker(
 func (p *PreFlight) CheckImmutablesDiffs(d r3diff.Changelog, diffChecker diffs.Checker) error {
 	var errs []error
 
-	r, err := rules.NewEKSClusterRulesExtractor(p.paths.DistroPath)
+	r, err := rules.NewEKSClusterRulesExtractor(p.paths.DistroPath, diffChecker.GetCurrentConfig())
 	if err != nil {
 		if !errors.Is(err, rules.ErrReadingRulesFile) {
 			return fmt.Errorf("error while creating rules builder: %w", err)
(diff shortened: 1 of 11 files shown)


ralgozino (approved):
lgtm 🚀

Issue #591, opened 2025-04-22
Feature: furyctl cmd <namecommand> <arguments...>


nutellinoit:
**Is your feature request related to a problem? Please describe.**
Since we download all the binaries needed by furyctl executions, it should also be possible to execute the correct binary using a furyctl subcommand.

**Describe the solution you'd like**
The feature I would like to be implemented is to add a command like:

```bash
furyctl cmd <namecommand> <arguments...>
```

For example something like:

```bash
furyctl cmd kubectl get pods
```

Should execute the correct kubectl version defined by the distribution version kfd.yaml file that has been downloaded in the `.furyctl/bin/` directory

ralgozino:
I really like this proposal! It is specially useful for managing plugins, for example you can use the right version of kapp or helm to uninstall a plugin and you don't need to install the tool.

I'd suggest `furyctl exec <tool> <args>` instead of `furyctl cmd`.

PR #618, opened 2025-06-26
feat: add `furyctl get aliases` command for tool management

nutellinoit:
### Summary 💡

Adds `furyctl tools` command supporting multiple shell integration formats (aliases, functions, mise).

Closes: #591

### Description 📝

**Command Structure:**
- `furyctl tools aliases` - Generate bash aliases 
- `furyctl tools functions` - Generate bash functions (higher precedence than aliases)
- `furyctl tools mise` - Generate/update mise.toml configuration

**Core Features:**
- Strict configuration validation (fails without furyctl.yaml)
- SD-based version selection from kfd.yaml
- Automatic dependency download (no manual steps required)
- Project-specific tooling with `--outdir` flag

**Implementation:**
1. Validates furyctl configuration exists
2. Downloads distribution and dependencies automatically
3. Parses tool versions from distribution's kfd.yaml  
4. Generates appropriate output format
5. Suppresses INFO logs for clean eval output

**Usage Examples:**
```bash
# View aliases
furyctl tools aliases --outdir $PWD

# Set functions (override version managers)
eval "$(furyctl tools functions --outdir $PWD)"

# Generate mise configuration
furyctl tools mise --outdir $PWD
```

**Key Technical Changes:**
- Shared tool discovery logic in `cmd/tools/shared.go`
- TOML parsing for mise integration using `github.com/pelletier/go-toml/v2`
- Concurrent map access fixes in configuration loading
- Functions provide higher shell precedence than aliases or version managers
- Project-specific vs global tooling patterns

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Unit tests for all three output formats
- [x] Integration testing with mise workflows
- [x] Linting and build verification
- [x] Manual testing with various flag combinations

### DIFF
--- a/cmd/tools/shared.go
+++ b/cmd/tools/shared.go
@@ -140,7 +140,6 @@ func DiscoverTools(flags SharedFlags) ([]ToolInfo, error) {
 
 		// Download the distribution.
 		res, err := distrodl.Download(flags.DistroLocation, furyctlPath)
-
 		if err != nil {
 			return nil, fmt.Errorf("error while downloading distribution: %w", err)
 		}
@@ -170,7 +169,6 @@ func DiscoverTools(flags SharedFlags) ([]ToolInfo, error) {
 		}
 
 		res, err := distrodl.Download(flags.DistroLocation, furyctlPath)
-
 		if err != nil {
 			return nil, fmt.Errorf("error while downloading distribution: %w", err)
 		}


cmd/tools/aliases.go
```diff
@@ -0,0 +1,34 @@
...
+
+// NewAliasesCmd creates the aliases command.
+func NewAliasesCmd() *cobra.Command {
+	return CreateShellIntegrationCommand(
+		"aliases",
+		"Generate bash aliases for downloaded tools",
+		`Generate bash aliases for tools downloaded by furyctl.
+
+This command scans the .furyctl/bin directory and outputs bash aliases that point
+to the downloaded tool binaries with their correct versions.
```
ralgozino, line 17: I think this is misleading, it scans the folder to check that the tools have been downloaded, but to the user what is important is that points to the right versions of the tools that are compatible with the distribution version defined in the furyctl.yaml

cmd/tools/mise.go
```diff
@@ -0,0 +1,385 @@
...
+
+	"github.com/sighupio/furyctl/internal/analytics"
+	"github.com/sighupio/furyctl/internal/app"
+	iox "github.com/sighupio/furyctl/internal/x/io"
+)
+
+const (
+	// FilePermissions represents the file permissions for mise.toml.
+	FilePermissions = 0o600
+)
```
ralgozino, line 26: there are some useful constants already defined in the x/io package:

"github.com/sighupio/furyctl/internal/x/io"

Also, `FilePermissions` is not a good name IMO, it should be the actual permissions like `OwnerReadWrite` or something.

cmd/tools.go
```diff
@@ -0,0 +1,33 @@
...
+		Short: "Manage tool integrations for downloaded binaries",
+		Long: `Manage tool integrations for downloaded binaries.
+
+Generate various output formats for integrating furyctl-downloaded tools 
+with your shell environment. Choose the format that works best with your 
+development setup:
+
+- aliases: Traditional bash aliases (default)
+- functions: Bash functions that override version managers like mise/asdf  
+- mise: Native mise integration via .mise.toml file`,
```
ralgozino, line 25:
```suggestion
- mise: Native mise integration via mise.toml file`,
```

cmd/tools/mise.go
```diff
@@ -0,0 +1,385 @@
...
+		false,
+		"Skip confirmation prompt when reverting tools",
+	)
+
+	return miseCmd
+}
+
+// updateMiseConfig creates or updates the mise.toml file with furyctl tools.
+func updateMiseConfig(tools []ToolInfo) error {
+	const miseFile = "mise.toml"
```
ralgozino, line 191: It would be nice if this was a flag instead of hard-coded. It may be that the furyctl.yaml and the mise.toml files are not in the same directory.

Having it as a flag would allow also user's that are still using `.mise.toml` and haven't migrated to `mise.toml` (without the dot) and one could even point to the global mise config (~/.config/mise/config.toml)

cmd/tools/mise.go
```diff
@@ -0,0 +1,385 @@
...
+	// Store all other sections to preserve them.
+	config.Other = rawConfig
+
+	return config, nil
+}
+
+// saveMiseConfig saves the mise configuration to file.
+func saveMiseConfig(filename string, config *MiseConfig) error {
+	// Combine tools section with other sections.
+	// This intermediate map is used to preserve the order of the sections in the final TOML file,
```
ralgozino, line 268: I don't think order is being preserved, my mise.toml gets sorted completed different after using the `tools mise` command, it even breaks some vars that depend on the previous one.

ralgozino: Also, it may be a good idea to run `mise fmt` after saving the file :)

cmd/tools/shared.go
```diff
@@ -0,0 +1,420 @@
...
+
+		// Validate base requirements.
+		depsvl := dependencies.NewValidator(executor, binPath, furyctlPath, false)
+		if err := depsvl.ValidateBaseReqs(); err != nil {
+			return nil, fmt.Errorf("error while validating requirements: %w", err)
+		}
+
+		// Download the distribution (suppress logs for clean output).
+		originalLevel := logrus.GetLevel()
+		logrus.SetLevel(logrus.ErrorLevel)
```
ralgozino, line 143: why? If I set the log level to DEBUG I want to see debug logs, please don't change it

cmd/tools/shared.go
```diff
@@ -0,0 +1,420 @@
...
+		logrus.SetLevel(originalLevel)
+
+		if err != nil {
+			return nil, fmt.Errorf("error while downloading distribution: %w", err)
+		}
+
+		basePath := path.Join(outDir, ".furyctl", res.MinimalConf.Metadata.Name)
+
+		// Validate the furyctl.yaml file (suppress logs for clean output).
+		logrus.SetLevel(logrus.ErrorLevel)
```
ralgozino, line 156: same as above

README.md
```diff
@@ -5,7 +5,7 @@
    <p>The Swiss Army Knife<br/>for the SIGHUP Distribution</p>
 
    [![Build Status](https://ci.sighup.io/api/badges/sighupio/furyctl/status.svg?ref=refs/heads/main)](https://ci.sighup.io/sighupio/furyctl)
-   ![Release](https://img.shields.io/badge/furyctl-v0.32.4-blue)
+   ![Release](https://img.shields.io/badge/furyctl-v0.32.5-blue)
```
ralgozino, line 8: Why is this showing up? May be this branch needs a rebase?

cmd/get.go
```diff
@@ -13,7 +13,7 @@ import (
 func NewGetCmd() *cobra.Command {
 	getCmd := &cobra.Command{
 		Use:   "get",
-		Short: "Get the kubeconfig, available upgrade paths for a cluster or compatible versions to use between SD, providers, furyctl",
+		Short: "Get the kubeconfig, available upgrade paths for a cluster, or compatible versions",
```
ralgozino, line 16: unwanted change?

ralgozino (changes requested):
I left some comments, the mise integration is not working for me but could be something on my end, I'll investigate and report back.

README.md
nutellinoit: Fixed with a new rebase

cmd/tools/shared.go
nutellinoit: fixd

cmd/tools.go
nutellinoit: Fixd

cmd/tools/mise.go
nutellinoit: Done ✅

cmd/tools/aliases.go
nutellinoit: Fixed

cmd/tools/shared.go
```diff
@@ -0,0 +1,396 @@
...
+			distrodl = dist.NewDownloader(client, typedGitProtocol, "")
+		}
+
+		res, err := distrodl.Download(flags.DistroLocation, furyctlPath)
+		if err != nil {
+			return nil, fmt.Errorf("error while downloading distribution: %w", err)
+		}
+
+		distroManifest = res.DistroManifest
+	}
```
Filo01, line 177: this big if/else block could probably be refactored in a better way with proper functions that say what they are doing

cmd/tools_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+		case "functions":
+			assert.Equal(t, "Generate bash functions for downloaded tools", subcmd.Short)
+			assert.Contains(t, subcmd.Long, "bash functions")
+			assert.Contains(t, subcmd.Long, "Functions have higher")
+
+		case "mise":
+			assert.Equal(t, "Generate or update mise.toml with downloaded tool paths", subcmd.Short)
+			assert.Contains(t, subcmd.Long, "mise.toml")
+		}
```
Filo01, line 60: testing for descriptions seems overly verbose, do we really need it?

PR #632, opened 2025-10-29
feat: add OpenTofu

stefanoghinelli:
### Summary 💡

Add OpenTofu support with deprecation warning for Terraform. Backward compatibility with previous SD releases is preserved.

Relates: [sighupio/distribution/pull/468](https://github.com/sighupio/distribution/pull/468) - [sighupio/product-management/issues/657](https://github.com/sighupio/product-management/issues/657)

### Description 📝

Starting from SD v1.33.2, distribution ships with OpenTofu instead of Terraform. This PR allows furyctl to work with both previous versions (Terraform only) and new ones (OpenTofu only). During the deprecation period, users can keep spec.toolsConfiguration.terraform in their furyctl.yaml without blocking errors, but under the hood furyctl will use the tofu binary.

Changelog:

- New OpenTofu tool: added internal/dependencies/tools/opentofu.go with download, validation, and version detection
- Feature detection: added FeatureOpentofuSupport in internal/distribution/features.go (checks kfd.yaml for tools.common.opentofu.version)
- Binary selection (internal/cluster/phase.go): uses tofu binary when opentofu.version is set in kfd.yaml, otherwise terraform
- Preflight state handling (internal/apis/kfd/v1alpha2/ekscluster/common/preflight.go): added helper to handle both spec.toolsConfiguration.terraform and spec.toolsConfiguration.opentofu state configs
- Validation (internal/config/validate.go): shows deprecation warning when user has spec.toolsConfiguration.terraform and opentofu is available
- Upgrade paths (configs/upgrades/ekscluster/1.33.1-1.33.2/): added pre-phase scripts to backup terraform states before upgrading with OpenTofu

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Upgrade from SD v1.33.1 to v1.33.2-rc.0 keeping terraform config, warning displayed correctly
- [x] New SD cluster with v1.33.2-rc.0 using opentofu field, with no warnings
- [x] New SD cluster with v1.33.2-rc.0 using terraform field, warning displayed
- [x] Upgrade from SD v1.33.1 to v1.33.2-rc.0 migrating to opentofu

### Future work 🔧

In the future versions (likely after v1.35), we'll need to change the Terraform deprecation warning to a blocking error.

### DIFF
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -400,7 +400,7 @@ func validateToolsConfiguration(repoPath string, furyctlConf map[string]any) err
 
 	toolsConfig, exists := spec["toolsConfiguration"].(map[string]any)
 	if !exists {
-		// No error if there is no toolsConfiguration(e.g. not available in KFDDistribution)
+		// No error if there is no toolsConfiguration(e.g. not available in KFDDistribution).
 		return nil
 	}
 


nutellinoit (changes requested):
LGTM with just one issue

kriive (approved):

internal/apis/kfd/v1alpha2/ekscluster/common/preflight.go
```diff
@@ -50,6 +50,19 @@ type PreFlight struct {
 	InfrastructureTerraformOutputsPath string
 }
 
+func (p *PreFlight) getTerraformStateConfig() private.SpecToolsConfigurationTerraformStateS3 {
+	if p.FuryctlConf.Spec.ToolsConfiguration.Opentofu != nil {
+		return private.SpecToolsConfigurationTerraformStateS3{
```
ralgozino: just curious: why do you need to return a new struct? can't you return here `p.FuryctlConf.Spec.ToolsConfiguration.Opentofu.State.S3` here like below for terraform?

ralgozino: that makes sense, thanks!

PR #649, opened 2026-03-03
fix(cmd): validate positional args

### LINKED ISSUE #648: `furyctl apply` silently ignores positional arguments
### Describe the bug

The `furyctl apply` command silently ignores positional arguments and failing highlight potential issues to the user.

Example:

```
$ furyctl apply distribution
```

Will apply all the phases and should instead show a message to the user saying that `distribution` is not a valid argument. Notice that the `--phase (-p)` flag is missing.

### To Reproduce

1. Run `furyctl apply <something>`
2. See how the cli runs the apply command and ignores the positional argument

### Expected behavior

The CLI should stop and tell the user that the positional argument is an error.

### Software versions

- `furyctl version` command output, for example:

```bash
$ furyctl version
buildTime: 2026-02-24T08:25:40Z
gitCommit: cc3e84c6110a4b5344516a7a0387a5349c23ee8d
goVersion: go1.23.2
osArch: arm64
version: 0.34.0
```

### Additional context

This issue is probably present on other commands

ralgozino:
Fail when user passes positional args to commands that do not require them, avoiding mistakes.

### Summary 💡

Modify all commands that don't take a positional argument to fail with an error when the user passes one (most probably by mistake).

Fixes #648

### Description 📝

See the linked issue on the rationale behind this change.

### Breaking Changes 💔

None, but furyctl invocations that included a positional argument by mistake could start failing now.

See for example the fix to the e2e test that were passing the `true` value to a flag that does not take one and it was being silently ignored as a positional argument.

### Tests performed 🧪

- [x] Tested that commands now fail with an error when a positional argument is passed instead of silently ignoring them.
- [x] Tested that commands still work as expected when they are invoked properly.

### Future work 🔧

None

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -78,12 +78,14 @@ func NewApplyCmd() *cobra.Command {
 	var cmdEvent analytics.Event
 
 	applyCmd := &cobra.Command{
+		Args:  cobra.NoArgs,
 		Use:   "apply",
 		Short: "Apply the configuration to create, update, or upgrade a battle-tested SIGHUP Distribution cluster",
 		Example: `  furyctl apply                                     Apply all the configuration to the cluster using the default configuration file name
   furyctl apply --config mycluster.yaml             Apply a custom configuration file
   furyctl apply --phase distribution                Apply a single phase, for example the distribution phase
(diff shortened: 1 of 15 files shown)


stefanoghinelli (approved):
Awesome! 🚀

PR #652, opened 2026-04-08
feat: add furyctl get cluster-info command

stefanoghinelli:
### Summary 💡

Implements the `furyctl get cluster-info` command, which displays a quick overview of a running cluster's configuration and status.

Closes: sighupio/product-management/issues/642

### Description 📝

All values are read from the configmaps/secrets that furyctl stores in the kube-system namespace; the Kubernetes version and node list are retrieved via live `kubectl` queries.

The last configuration change time is read from the `furyctl-config` secret. It is printed in UTC.

The ongoing upgrade status is read from the `furyctl-upgrade-state` configmap and reports the first pending or failed phase with its status.

Nodes are grouped by their primary `node-role.kubernetes.io/*` label. Since a node can carry multiple role labels, is applied a deterministic order: `control-plane` first, then `master`, then alphabetical, then those with no role.

The `--format` / `-f` flag supports `text` (default), `json`, and `yaml`.

#### Example txt output

```
./furyctl get cluster-info
Cluster Name:               test-onprem
SD Version:                 v1.34.0
SD Kind:                    OnPremises
SD Installer version:       v1.34.4
SD Upgrade paths:           None
SD Ongoing upgrade:         None
Kubernetes version:         v1.34.4
Etcd topology:              Stacked
Last configuration change:  2026-04-08 06:49:27 (UTC)
Custom Patches present:     Yes

Module             Version  Type
--------------------------------
Networking         v3.1.0   cilium
Ingress            v5.0.0   haproxy/single
Monitoring         v4.1.0   none
Logging            v5.3.0   none
Tracing            v1.4.0   none
Policy             v1.16.0  none
Auth               v0.6.1   none
Disaster Recovery  v3.3.0   none

Plugin Name             Type
----------------------------
local-path-provisioner  Kustomize
ldap-local              Kustomize
s1-agent                Helm

Node Role      Qty  vCPU  RAM(GiB)
---------------------------------
control-plane  1    2     3.8
infra          3    12    16.9
worker         2    8     11.3
<none>         1    4     5.6
=================================
Total          7    26    37.6
```

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Unit tests
- [x] Integration test
- [x] Linting passes with no issues
- [x] Manual test against all providers

### Future work 🔧

- This first implementation extracts everything is already available, which phase failed/is pending, and its status. The "from which version" detail cannot be fulfilled with the current data since the information simply isn't stored there. To support it, the upgrade.State struct would need to be extended to include From and To version fields, populated when furyctl starts an upgrade. That's a separate change in the upgrade flow, outside the scope of this PR.
- The “nice-to-have” features will be addressed in a future iteration.

### DIFF
--- a/cmd/get/cluster-info.go
+++ b/cmd/get/cluster-info.go
@@ -1,0 +1,360 @@
+// Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package get
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	v3 "gopkg.in/yaml.v3"
+
+	"github.com/sighupio/furyctl/internal/analytics"
+	"github.com/sighupio/furyctl/internal/app"
+	"github.com/sighupio/furyctl/internal/clusterinfo"
+	"github.com/sighupio/furyctl/internal/flags"
+	cobrax "github.com/sighupio/furyctl/internal/x/cobra"
+	execx "github.com/sighupio/furyctl/internal/x/exec"
+)
+
+const (
+	outputFormatText = "text"
+	outputFormatJSON = "json"
+	outputFormatYAML = "yaml"
+	minTableLines    = 3
+)
+
+var errInvalidOutputFormat = errors.New("invalid output format, supported values are: text, json, yaml")
+
+func NewClusterInfoCmd() *cobra.Command {
+	var cmdEvent analytics.Event
+
+	clusterInfoCmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "cluster-info",
+		Short: "Display cluster information.",
+		Long:  `Display information about the cluster, Kubernetes, and SD status. The command provides a quick overview of the cluster configuration and its current state; its output can be used for analysis and troubleshooting.`,
+		Example: `  furyctl get cluster-info                display cluster info in text format (default)
+  furyctl get cluster-info --format json   display cluster info as JSON
+  furyctl get cluster-info --format yaml   display cluster info as YAML
+ `,
+		PreRun: func(cmd *cobra.Command, _ []string) {
+			cmdEvent = analytics.NewCommandEvent(cobrax.GetFullname(cmd))
+
+			if err := flags.LoadAndMergeCommandFlags("get"); err != nil {
+				logrus.Fatalf("failed to load flags from configuration: %v", err)
+			}
+
+			if err := viper.BindPFlags(cmd.Flags()); err != nil {
+				logrus.Fatalf("error while binding flags: %v", err)
+			}
+		},
+		RunE: func(_ *cobra.Command, _ []string) error {
+			ctn := app.GetContainerInstance()
+
+			tracker := ctn.Tracker()
+			tracker.Flush()
+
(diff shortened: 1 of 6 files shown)


internal/clusterinfo/collector.go
```diff
@@ -0,0 +1,763 @@
...
+		}
+	}
+
+	return false
+}
+
+// extractModules combines types from furyctl config with versions from the KFD
+// YAML. Fixed order keeps output stable; for EKS, an AWS row is appended when
+// present.
+func extractModules(configMap map[string]any, sd distroconf.KFD, kind string) []ModuleInfo {
```
stefanoghinelli, line 391: Note for reviewers: `extractModules` is intentionally static, display names differ from yaml keys (opa is "Policy", dr is "Disaster Recovery"), each module has its own type extraction logic, and versions are sourced from the typed distroconf.KFD struct. Adding a new module requires an explicit entry.

cmd/get/cluster-info.go
```diff
@@ -0,0 +1,360 @@
...
+
+func NewClusterInfoCmd() *cobra.Command {
+	var cmdEvent analytics.Event
+
+	clusterInfoCmd := &cobra.Command{
+		Args:  cobra.NoArgs,
+		Use:   "cluster-info",
+		Short: "Display cluster information.",
+		Long:  `Display information about the cluster, Kubernetes, and SD status. The command provides a quick overview of the cluster configuration and its current state; its output can be used for analysis and troubleshooting.`,
+		Example: `  furyctl get cluster-info                display cluster info in text format (default)
```
ralgozino: columns alignment
```suggestion
		Example: `  furyctl get cluster-info                 display cluster info in text format (default)
```

cmd/get/cluster-info.go
```diff
@@ -0,0 +1,360 @@
...
+			"When set, furyctl looks for kubectl inside this folder. "+
+			"If not set, kubectl is resolved from PATH.",
+	)
+
+	clusterInfoCmd.Flags().StringP(
+		"format",
+		"f",
+		outputFormatText,
+		"Output format. Supported values: text, json, yaml",
+	)
```
ralgozino, line 124: It would be nice to autocomplete the available options here like we do for the other commands. See for example the apply.go file and search for "completion"

internal/clusterinfo/collector.go
```diff
@@ -0,0 +1,763 @@
...
+	}
+
+	if !hasAny {
+		return nil
+	}
+
+	return result
+}
+
+func etcdTopology(kind string, configMap map[string]any) string {
```
ralgozino, line 525: this func will need to be adapted to support immutable

internal/clusterinfo/collector.go
```diff
@@ -0,0 +1,763 @@
...
+
+	hosts, ok := etcd["hosts"].([]any)
+	if !ok || len(hosts) == 0 {
+		return etcdStacked
+	}
+
+	return etcdDedicated
+}
+
+func installerVersion(kind string, sd distroconf.KFD) string {
```
ralgozino, line 604: this func will need to be adapted to support immutable

ralgozino (commented):
A first round of review, I'll take a closer look in the next days

internal/clusterinfo/collector.go
stefanoghinelli: Cannot be added now because fury-distribution v1.34.0 does not yet expose KFDKubernetes.Immutable. I will add the case along with the dependency update once Kubernetes.Immutable is merged into the main branch. WDYT?

ralgozino: This can be addressed now

stefanoghinelli:
Ready @ralgozino

ralgozino (approved):
lgtm, very nice addition <3

nutellinoit (approved):
LGTM! Ship it! 🚀

PR #656, opened 2026-04-30
fix: fallback to spec.kubernetes.vpcId when infra output.json has no vpc_id

mimnix:
### Summary 💡

   When running the infrastructure phase without creating a VPC (because it already exists), the `output.json` file is created but does not contain a `vpc_id` key. This causes `ErrVpcIDNotFound` in the distribution phase, preventing the fallback to `.spec.kubernetes.vpcId` from the furyctl config.

   Closes:

   Relates:

   ### Description 📝

   The `extractVpcIDFromPrevPhases` function in the distribution phase had an `if/else` structure where the fallback to `.spec.kubernetes.vpcId` was only reached when `output.json` did not exist at all (`os.ReadFile` error).

   When the infrastructure phase runs without creating a VPC, it still produces an `output.json` (empty or without `vpc_id`). In this case, the code entered the `if` branch, found no `vpc_id`, and returned `ErrVpcIDNotFound` instead of falling through to the furyctl config fallback.

   The fix removes the exclusive `if/else` structure so that when `output.json` exists but does not contain `vpc_id`, execution continues to the fallback that reads the value from `.spec.kubernetes.vpcId`.

   ### Breaking Changes 💔

   None.

   ### Tests performed 🧪

   - [x] Project compiles successfully with `go build ./...`
   - [ ] Tested the distribution phase with an existing VPC and no `vpc_id` in infra output

   ### Future work 🔧

   - Add unit tests for `extractVpcIDFromPrevPhases` covering the different fallback scenarios.
   - Consider applying the same fix to `kubernetes.go` where `ErrVpcIDNotFound` is also returned in a similar pattern.

### DIFF
--- a/internal/apis/kfd/v1alpha2/ekscluster/common/distribution.go
+++ b/internal/apis/kfd/v1alpha2/ekscluster/common/distribution.go
@@ -142,34 +142,30 @@ func (d *Distribution) extractVpcIDFromPrevPhases(fMerger *merge.Merger) (string
 		var infraOut terraform.OutputJSON
 
 		if err := json.Unmarshal(infraOutJSON, &infraOut); err == nil {
-			if infraOut["vpc_id"] == nil {
-				return vpcID, ErrVpcIDNotFound
+			if infraOut["vpc_id"] != nil {
+				vpcIDOut, ok := infraOut["vpc_id"].Value.(string)
+				if !ok {
+					return vpcID, ErrCastingVpcIDToStr
+				}
+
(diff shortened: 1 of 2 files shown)


stefanoghinelli (commented):
Hi @mimnix

I think we need to cover testing more thoroughly.

Can you provide more information about testing the changes?

stefanoghinelli:
As discussed, @marcopaggioro  will work the unit tests.

stefanoghinelli (approved):

PR #659, opened 2026-05-12
chore: cqi

marcopaggioro:
### Summary 💡

⚠️ To review after the #638  because it's a child branch.

This PR aims to do some Code Quality Improvement, even from the update of Go to .24 from the PR #638 

### Description 📝

Notes:
- Removed no more necessary `.tool-versions` file
- Removed a "useless" struct created only to run `Sort` function in favor of `SortFunc`
- Replaced the combination of `HasPrefix` + `TrimPrefix` with `CutPrefix`
- Better error joining with `errors.Join` that preserve error types
- `slices.Contains` and `slices.ContainsFunc` instead of a manual for-loop
- Use the testkit `t.Chdir` and `t.TempDir` to handle directories for testing
- Use the higher level function `require.Equal` instead of `reflect.DeepEqual`
- Use testkit `t.Context()` instead of background context

### Breaking Changes 💔

Not detected.

### Tests performed 🧪

Ran all test suites.

### Future work 🔧

Not planned.


internal/flags/expansion_integration_test.go
```diff
@@ -171,13 +171,7 @@ func TestFlagExpansion_LoadFromCurrentDirectory(t *testing.T) {
...
 
 	// Change to temp directory
-	originalDir, _ := os.Getwd()
-	defer func() {
-		os.Chdir(originalDir)
-	}()
-
-	err = os.Chdir(tempDir)
-	require.NoError(t, err)
+	t.Chdir(tempDir)
```
ralgozino, line 174: From the docs I see that `t.Chdir` also restores to the previous directory, nice. But what happens in the case of an error? For example, if tempDir does not exist? I don't see `Chdir` returning an err.

.tool-versions
ralgozino, line 1: I don't see that all the tools specified in this file are also present in the mise.toml file. Is that correct or did we miss to port them to the mise.toml file?

internal/flags/expansion_integration_test.go
marcopaggioro: The function does not return an error, it throws panic errors or c.Fatal errors: in either cases the test fails.

.tool-versions
marcopaggioro: They seems unused. Local build and local testing works fine without them.
We can re-add them in the future if necessary.

ralgozino (approved):
lgtm

PR #671, opened 2026-06-16
feat: self-contained ansible bundle and per-kind dependency download

nutellinoit:
### Summary 💡

Three related improvements:
1. **Ansible becomes a downloadable, self-contained bundle** (relocatable CPython + ansible-core + collections) → no host Ansible for OnPremises/Immutable.
2. **Integrity**: when the distribution pins per-arch checksums, furyctl verifies the downloaded artifact (SHA-256) and rejects a mismatch (fail-closed).
3. furyctl now downloads/validates **only the dependencies each cluster kind needs** (across `apply`, `download dependencies` and validation).

Related distribution PR: https://github.com/sighupio/distribution/pull/525 (requires `fury-distribution v1.34.2-rc.1`)
Reference PoC (bundle build recipe + releases): https://github.com/nutellinoit/ansible-portable-poc (release `v0.2.1`)

### Description 📝

**Self-contained Ansible bundle**
- `internal/dependencies/tools/ansible.go`: `SupportsDownload()=true`; URL built from the pinned bundle release; **presence-only** version check (the pin is the bundle tag, not the ansible-core version).
- `internal/tool/ansible/runner.go`: runs `python3 <ansible-playbook>` (ignores the build-time shebang after relocation) + injects `ANSIBLE_COLLECTIONS_PATH` / `PATH`; falls back to system Ansible when not pinned.
- **Upgrade scripts** (`configs/upgrades/onpremises/**/{pre-kubernetes,pre-distribution}.sh.tpl`) call `{{ $.paths.ansiblePlaybook }}` instead of a bare `ansible-playbook`, so upgrades also use the bundle (or system Ansible when not pinned).
- Feature-gated by `tools.common.ansible.version`.

**Checksum verification**
- `pkg/dependencies/download.go`: `DownloadTools` appends go-getter `?checksum=sha256:<hash>` when the distribution pins a per-arch checksum (`KFDTool.Checksums`, keyed `<os>-<arch>`). Mismatch → the download fails. Tools/distros without checksums are unaffected.

**Per-kind dependency download**
- New `internal/distribution.ToolNeededForKind / ModuleNeededForKind / InstallerNeededForKind`, threaded through `DownloadAll/Installers/Tools/Modules` and the tools validator.
- OnPremises no longer pulls `terraform`/`awscli` or the `eks`/`immutable` installers.
- Adopts the `installer-<name>` repo naming convention (fixes the immutable installer URL); supports the new `Immutable` kind (which also uses Ansible).

### Breaking Changes 💔

None — without `tools.common.ansible.version` the system-Ansible path is unchanged; checksum verification is skipped when no checksum is pinned; per-kind filtering only narrows downloads to what each kind already used.

### Transition / backward compatibility 🔁

- Distros that **do not** pin `tools.common.ansible.version` (all existing ones) keep using the **system Ansible** — a host Ansible is still required there. The bundle kicks in only once a distro pins it.
- Maintainer note: **new** on-prem upgrade scripts that invoke Ansible must use `{{ $.paths.ansiblePlaybook }}` (root context, since the calls live inside `{{ range }}` blocks), not a bare `ansible-playbook`.

### Tests performed 🧪

- [x] `mise run lint` (0 issues), unit tests (incl. the checksum helper), `go build ./...`
- [x] On-prem e2e on Vagrant (1 master / 1 haproxy / 1 worker): `apply --phase kubernetes` creates a v1.34.4 cluster using the bundled Ansible
- [x] **Checksum**: cold-cache download verifies the pinned SHA-256; a wrong hash fails the download (no extraction)
- [x] `download dependencies` fetches only the onpremises installer + ansible bundle (no terraform/awscli/eks/immutable)

### Next steps 🔜

- **Internalize the PoC** into a SIGHUP-hosted bundle repo and make the bundle base URL **configurable per distro** (not hardcoded), with a CI check that rejects non-`sighupio` URLs.
- **Version validation**: allowlist `tools.common.ansible.version` (reject `/`, `\`, `..`) at all construction sites.
- Optional: cosign/minisign **signature** on top of the checksum.
- **macOS control node**: strip `com.apple.quarantine` (and/or ad-hoc re-sign) the extracted bundle so Gatekeeper doesn't block the bundled `python3` on Apple Silicon. Linux control nodes (the on-prem/CI case) are unaffected.


nutellinoit:
Superseded by mise bundle, closing

PR #679, opened 2026-06-25
feat(onpremises): enforce reducer-less unsupported rules + run migration playbooks in the kubernetes phase

nutellinoit:
### Summary 💡

Let the rules engine enforce `unsupported` transitions on rules without a reducer, and run migration playbooks from the onpremises kubernetes phase, so a config change is actually applied to a running cluster, not just validated. Enables the kube-proxy `ipvs/nil → nftables` migration (distribution side: sighupio/distribution#505).

### Description 📝

The `unsupported` check only looked at rules with a reducer, so a rule with only `unsupported` (like `kubeProxy.type`) was never enforced, forcing a fake reducer. Now `GetUnsupportedRules` feeds the preflight check in every provider, independent of reducers. It also exports `diffs.ExpandMapChanges` so `nil → value` transitions (a new field whose parent is born in the diff) still match leaf-targeted rules.

It also extends the reducer mechanism, until now distribution-only (`scripts/<lifecycle>.sh`), to the onpremises kubernetes phase, where it runs `migrations/<lifecycle>.yaml` via the ansible runner with the reducers' `from`/`to` as extra-vars. The distribution drives the migration entirely from the rules, no furyctl changes. As in the distribution phase, it prompts for confirmation before an unsafe migration (`--force migrations` to skip).

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Unit tests (`pkg/diffs`, `pkg/reducers`) incl. `nil → value`; full suite green, lint clean.
- [x] e2e on a real onprem 1.35 cluster: `ipvs → nftables` applied (kube-proxy + Calico on nftables), idempotent re-apply, `→ ipvs`/`→ none` blocked.
- [x] Confirmation prompt verified e2e.

### Future work 🔧

None.

### DIFF
--- a/internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
+++ b/internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
@@ -373,15 +373,15 @@ func (p *PreFlight) CheckReducersDiffs(d r3diff.Changelog, diffChecker diffs.Che
 
 	errs = append(errs, diffChecker.AssertReducerUnsupportedViolations(
 		d,
-		r.UnsupportedReducerRulesByDiffs(r.GetReducers("infrastructure"), d),
+		r.GetUnsupportedRules("infrastructure"),
 	)...)
 	errs = append(errs, diffChecker.AssertReducerUnsupportedViolations(
 		d,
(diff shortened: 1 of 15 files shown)


stefanoghinelli (approved):
Super!

PR #681, opened 2026-06-25
feat: manage ansible (+ galaxy collections) via the bundled mise

nutellinoit:
### Summary 💡

`furyctl` installs Ansible through the bundled mise (`pipx:ansible-core` on a mise-managed `uv` + `python`) plus the distribution-pinned galaxy collections, for OnPremises/Immutable — instead of requiring Ansible on the host. Fully gated: when the kind's provider section (`tools.onpremises` / `tools.immutable`) does not pin ansible, it stays a host dependency, exactly as today.

Closes:

Relates: https://github.com/sighupio/distribution/pull/539

### Description 📝

Ansible is pinned per provider (`tools.onpremises` / `tools.immutable`, mirroring `tools.eks`); `EffectiveAnsible(tools, kind)` resolves it for the running kind. When pinned, the generated `mise.toml` gets `pipx:ansible-core` plus a mise-managed `uv` and `python` (distro versions, furyctl defaults as fallback), and the collections are installed with `ansible-galaxy`. `uv` is needed because mise's pipx backend requires it (absent on a clean machine); `python` is needed because otherwise `uv` fetches its own interpreter outside `mise/data`, which wouldn't be packed into the air-gapped bundle.

Ansible is materialized as a single relative `venv` symlink and run as `venv/bin/python venv/bin/ansible-playbook` — going through the venv root preserves CPython's `pyvenv.cfg` detection, which a flattened symlink would break. When the kind doesn't pin ansible, the runner falls back to bare `ansible`/`ansible-playbook`. Upgrade scripts call `{{ $.paths.ansiblePlaybook }}` (the mise command when managed, bare otherwise), and the air-gap extraction repoints the bundled venv to the relocated python so it runs offline.

Building the bundle on a clean host also surfaced a pre-existing issue, fixed here: helm-diff's install hook calls `helm` by bare name and downloads via curl/wget, but furyctl drives tools by absolute path. `helmfile init` now puts helm on PATH and fails early with a clear error when curl/wget is missing — only when the plugin isn't already installed, so offline air-gap runs need neither.

### Breaking Changes 💔

None — the mise path activates only when the kind's provider section pins ansible; otherwise Ansible stays a host dependency.

> Maintainer note: new on-prem upgrade scripts that invoke Ansible must use `{{ $.paths.ansiblePlaybook }}` (root `$.` context, since the calls live inside `{{ range }}` blocks), not a bare `ansible-playbook`.

### Tests performed 🧪

- [x] `mise run format-go && mise run lint` (0 issues), `mise run test-unit`, `go build ./...`
- [x] On-prem e2e on Vagrant, furyctl run **inside a Linux container**: `apply` creates a v1.35.5 cluster (master+worker Ready), playbooks run by the mise-managed Ansible 2.21.0 + collections from `tools.onpremises`
- [x] Host fallback against a real old distro (v1.34.1, no ansible sections): no venv materialized, Ansible stays on the host; EKS/KFDDistribution never pull ansible
- [x] Clean Linux container (no host python/uv/pipx): self-contained install; `get kubeconfig` runs `ansible all -m ping` + fetches the kubeconfig over SSH
- [x] Air-gap on Linux, end to end: bundle built in a clean container, then consumed in a `--network none` container — the bundled Ansible + collections run fully offline
- [x] Bundle build with/without curl: clear error when missing, builds when present (no system helm required)

### Future work 🔧

- e2e coverage in CI for both the managed and host-fallback Ansible paths.
- Explore mise's OCI support to ship a distribution version's tools (incl. the ansible venv) from a single internal registry.


docs/releases/v0.35.0.md
```diff
@@ -10,12 +10,14 @@ Welcome to the latest release of `furyctl` maintained by SIGHUP by ReeVo team.
...
 - EKSCluster: restored the `json` struct tags on the hand-maintained config types introduced in #674. Without them the Terraform/OpenTofu data injection (VPC id, IAM role ARNs, Route53 hosted zone) was keyed by Go field names instead of the camelCase schema keys, so the distribution phase rendered an empty `vpc_id` and `tofu plan` failed on every EKS cluster.
 - [[#667](https://github.com/sighupio/furyctl/pull/667)] Fixed a panic that occurs when deploying an EKS cluster with spec.kubernetes.apiServer.publicAccess: false and using an existing VPC.
+- [[#681](https://github.com/sighupio/furyctl/pull/681)] Installing the helm plugins (helm-diff) now works on hosts without a system `helm` on the `PATH` (furyctl drives tools by absolute path), both when building the air-gapped bundle and during the plugins phase. It also fails early with a clear error when neither `curl` nor `wget` is available to download the plugin, but only when the plugin is not already installed, so offline air-gapped runs are unaffected.
```
ralgozino: the bug related to this fix was never released, I think we can skip it from the release notes

ralgozino (approved):
One non-blocking observation. The rest lgtm

PR #685, opened 2026-06-29
feat: relative path for customResources

ralgozino:
### Summary 💡

Add support for a new `spec.distribution.customResources` field that lets users inject extra Kustomize resources (local files, local kustomize bases, or remote git/URL references) into the distribution phase's main `kustomization.yaml`.

furyctl rewrites local resource paths so they are relative to the rendered manifests directory before templating, because `kustomize build` rejects absolute paths in its `resources` list (`new root ... cannot be absolute`).

Closes:

Relates:
- https://github.com/sighupio/distribution/pull/543

### Description 📝

`customResources` is a first-class sibling of `customPatches` under `spec.distribution` (I originally consider adding new field under `customPatches` but resources aren't patches, so they don't belong there). It is a flat array of strings, mirroring kustomize's own top-level `resources:` list:

```yaml
spec:
  distribution:
    customResources:
      - ./gapi-calico                          # local base, relative to furyctl.yaml
      - ../../plugins/kustomize/localstorage   # local base outside the project (warns)
      - github.com/org/repo//modules/x?ref=v1  # remote ref
```

The furyctl-side change is a single post-mapping step in the template engine, `relativizeCustomResources` in `pkg/template/model.go`, called in `Generate()` right after `MapDynamicValuesAndPaths()`. For each entry of `spec.distribution.customResources`:

- Local file/dir (resolves to an existing path) → rewritten to a path relative to `<TargetPath>/manifests` via `filepath.Rel`. The referenced files are left in place so kustomize bases keep composing with their own relative references (the distribution `apply.sh` already uses `--load-restrictor LoadRestrictionsNone`).
- Remote resource (git/URL) or non-existent path → left untouched so kustomize can fetch/resolve it itself.
- Path outside the furyctl config directory → still rewritten, but a warning is logged because the resulting relative path may differ between machines.

#### Design notes for reviewers

- The rewrite runs _after_ the mapper on purpose: the mapper has already turned the user's `./foo` inputs into absolute paths, and `TargetPath` (the render output location) is only known on the Model, never in the template context — so a custom template function could not have computed this.
- Local-vs-remote is classified with `os.Stat` rather than by parsing kustomize's git/URL grammar: existing paths are rewritten, everything else passes through.
- Emitting a path relative to the config directory (instead of an absolute one) keeps the rendered output deterministic across machines. Is up to the user to check those folder into git if applicable.
- I also considered copying the whole target directory into a `custom-resources` folder near the rest of the manifests inside the outdir, like we do for the `vendor`ed modules. This approach has several downsides:
  1. It "breaks" when the target folder relies on other folders as bases/resources. We would need to consider that and parse somehow to detect it and copy that too.
  2. We could have name collisions, so we would have to add some kind of hash that probably would've changed between machines.
  3. We could run a kustomize build of that folder and save it into an "`out.yaml`" file but it did not feel right.

### Breaking Changes 💔

None, new optional field.

### Tests performed 🧪

- [x] Tested adding custom resources: local files, local kustomize folders, remote files, remote git repos
- [x] Tested apply still works not setting the new field

### Future work 🔧

None

### DIFF
--- a/pkg/template/model.go
+++ b/pkg/template/model.go
@@ -12,12 +12,18 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
 
 	iox "github.com/sighupio/furyctl/internal/x/io"
 	"github.com/sighupio/furyctl/pkg/template/mapper"
 	yamlx "github.com/sighupio/furyctl/pkg/x/yaml"
 )
+
+// distributionManifestsDir is the subdirectory of the distribution render target
+// where the main kustomization.yaml is written. Entries of
+// spec.distribution.customPatches.resources are resolved relative to it.
+const distributionManifestsDir = "manifests"
 
 var (
 	ErrTargetIsNotEmpty = errors.New("target directory is not empty")
@@ -128,6 +134,8 @@ func (tm *Model) Generate() error {
 	if err != nil {
 		return fmt.Errorf("error mapping dynamic values: %w", err)
 	}
+
+	tm.relativizeCustomResources(context)
 
 	tm.Context = context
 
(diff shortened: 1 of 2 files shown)


nutellinoit (approved):
LGTM!

PR #687, opened 2026-06-30
fix(template): warn once for custom resources outside config dir

ralgozino:
### Summary 💡

Warn only once the user when customResources are pointing outside the config dir.

Relates:
- #685 

### Description 📝

The "custom resource resolves outside the furyctl configuration directory" warning was emitted on every template render pass. Since the distribution templates are rendered several times during a single apply (preflight, preupgrade, distribution phase), the same warning showed up multiple times.

Track already-warned paths in a process-wide set so each out-of-project custom resource warns at most once per run, keyed by resolved absolute path so distinct resources still get their own warning.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested apply and verified that the warning shows up only once.

### Future work 🔧

None

> [!IMPORTANT]
>

~~- [ ] I've updated the `docs/releases/unreleased.md` file (or equivalent)~~
- [x] CI is green

### DIFF
--- a/pkg/template/model.go
+++ b/pkg/template/model.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	"github.com/sirupsen/logrus"
 	"gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


stefanoghinelli (approved):
lgtm!

PR #688, opened 2026-06-30
feat(immutable): add --upgrade for kind immutable 

smerlos:
### Summary 💡

Add immutable cluster **upgrade** support. furyctl renders the Ansible role version
variables from `immutable.yaml` (single source), validates the immutable config, and fixes a race in the
assets server. Immutable k8s is coupled to the SD minor (1.35.5), like `onpremises`.

Closes:

Relates: sighupio/distribution#544, sighupio/installer-immutable#8, installer-immutable tag `v1.35.5-rc.0`

### Description 📝

- `buildVersionVars` / `renderVersionVarsFile` / `selectImmutableAssets` — render `group_vars/all.yml` from
  `immutable.yaml` so roles get k8s / sysext / flatcar / haproxy / csr-approver / registry vars from one
  source (install == upgrade; unit-tested for determinism).
- `cmd/validate/immutable.go` — validate `kfd.immutable.version` exists in the vendored manifest.
- `cmd/serve/path.go` — mutex around `nodesStatus` (concurrent-map race during PXE boot).
- `configs/upgrades/immutable/1.34.8-1.35.5/` — upgrade-script payload (for the future upgrade work).

Reuses the release's existing immutable engine + installer-by-tag download — those are untouched.

### Breaking Changes 💔

None. Additive only.

### Tests performed 🧪

- [x] `go build ./...`, `go vet`, unit tests (version-vars determinism) — green; code lint clean
- [x] End-to-end install + upgrade 1.34.8 → 1.35.5 (control-plane first, then per-node workers) → 6/6,
      validated on an onprem-consistent setup
- [x] Config validated against the v0.35.0 release distro (`distributionVersion: v1.35.0`, immutable 1.35.5)
- [x] Fresh-install recreate against this exact branch — pending

### Future work 🔧

Immutable upgrade, onprem-style: trigger via the common `preupgrade.go` distributionVersion diff with
SD-named upgrade-path dirs; needs a prior SD release shipping immutable 1.34.x. Until then this is
install-only at the SD-coupled k8s version.

- [ ] I've updated the `docs/releases/unreleased.md` file (or equivalent)
- [x] I've considered all the different cluster kinds — this change is immutable-scoped and leaves OnPremises / EKSCluster / KFDDistribution untouched
- [x] CI is green


cmd/validate/config.go
```diff
@@ -124,6 +124,18 @@ func NewConfigCmd() *cobra.Command {
...
+			if res.MinimalConf.Kind == immutableKind {
+				if err := validateImmutableVersion(res, typedGitProtocol); err != nil {
+					logrus.Error(err)
+
+					cmdEvent.AddErrorMessage(ErrValidationFailed)
+					tracker.Track(cmdEvent)
+
+					return ErrValidationFailed
+				}
+			}
```
ralgozino: I don't get why are we doing this check and fast-fail. Is this a reflux from when you were using the installer version instead of the kubernetes version as key in the immutable.yaml?

configs/upgrades/immutable/1.34.8-1.35.5/post-distribution.sh.tpl
ralgozino, line 1: do we need this file?

configs/upgrades/immutable/1.34.8-1.35.5/pre-distribution.sh.tpl
ralgozino, line 1: do we need this file?

configs/upgrades/immutable/1.34.8-1.35.5/pre-kubernetes.sh.tpl
```diff
@@ -0,0 +1,20 @@
...
+
+## etcd upgrade: stage the etcd sysext + renew certificates, serial:1 (quorum preserved).
+ansible-playbook upgrade-etcd.yaml --become
+
+## control-plane upgrade: refresh the kubeadm CLI + kubeadm upgrade apply before the reboot, serial:1.
+ansible-playbook upgrade-control-plane.yml --become
+
+{{- if ne .upgrade.skipNodesUpgrade true }}
+## worker upgrade: drain -> stage -> reboot -> kubeadm upgrade node -> uncordon, serial:1.
+ansible-playbook upgrade-worker-nodes.yml --become
```
ralgozino: the logic for limit to a single node is missing here

internal/apis/kfd/v1alpha2/immutable/create/infrastructure.go
```diff
@@ -92,7 +92,7 @@ func (i *Infrastructure) Exec(_ string, upgradeState *upgrade.State) error {
 	i.CopyPathsToConfig(&mCfg)
 
 	mCfg.Data["kubernetes"] = map[any]any{
-		"version": i.kfdManifest.Kubernetes.Immutable.Version,
+		"version": normalizeVersion(i.kfdManifest.Kubernetes.Immutable.Version),
```
ralgozino: kubernetes version is set by us in the kfd.yaml, why would we need to sanitize it?

I would pass it as-is, because if we want use some other string, for example for development purposes, this will drop the initial `v` silently.

internal/apis/kfd/v1alpha2/immutable/create/infrastructure.go
```diff
@@ -114,6 +114,26 @@ func (i *Infrastructure) Exec(_ string, upgradeState *upgrade.State) error {
...
+	versionVarsTpl := filepath.Join(
+		i.paths.DistroPath, "templates", cluster.OperationPhaseKubernetes, "immutable", "group_vars", "all.yml.tpl",
+	)
+
+	if err := renderVersionVarsFile(
+		targetPath, normalizeVersion(i.kfdManifest.Kubernetes.Immutable.Version), i.KubectlPath,
+		versionVarsTpl, immutableAssets,
+	); err != nil {
+		return fmt.Errorf("error rendering version vars file: %w", err)
+	}
```
ralgozino: I'm not sure rendering a template file that belongs to the kubernetes phase in the infrastructure phase is a good idea.

If the file is actually needed in several phases I think it would be better to have it inside each phase directory or in some common place. Because, if someone modifies a template file inside the `kubernetes` folder they will be impacting other phases without knowing it.


I might be wrong, but it seems to me that none of the variables in the `group_vars/all.yaml` file are used by the roles executed in the infrastructure phase. So there's no need for it.

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
```diff
@@ -262,6 +268,113 @@ func (i *Infrastructure) getImmutableAssets() (assets, error) {
...
+	}
+
+	var manifest immutableManifest
+	if err := yaml.Unmarshal(data, &manifest); err != nil {
+		return assets{}, fmt.Errorf("error parsing immutable manifest: %w", err)
+	}
+
+	if len(manifest.Kubernetes) == 0 {
+		return assets{}, ErrNoKubernetesVersions
+	}
```
ralgozino: I don't think this check is really needed, we won't have an immutable.yaml file with no kubernetes versions defined. But better safe than sorry, we can keep it.

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
ralgozino, line 1: if we don't actually need the group_vars/all.yaml file in the infrastructure phase all these changes are not needed any more.

ralgozino (changes requested):

internal/apis/kfd/v1alpha2/immutable/create/infrastructure.go
smerlos: Removed it — the version comes from `kfd.yaml` and never has a leading `v`, so the `TrimPrefix` only ever risked silently rewriting the key. Passing it as-is now; a bad version just fails loudly with `ErrKubernetesVersionNotFound`.

cmd/validate/config.go
smerlos: Dropped it (and `immutable.go`). It re-checked the k8s version that `selectImmutableAssets` already checks at create time, and paid an extra `installer-immutable` clone to do it. Single check now.

configs/upgrades/immutable/1.34.8-1.35.5/post-distribution.sh.tpl
smerlos: Deleted both — they only `echo`'d, and `Upgrade.Exec` already skips a missing hook. We'll add a real one when a transition actually needs it, like on-prem.

configs/upgrades/immutable/1.34.8-1.35.5/pre-distribution.sh.tpl
smerlos: Same — deleted (see the note on `post-distribution.sh.tpl`).

configs/upgrades/immutable/1.34.8-1.35.5/pre-kubernetes.sh.tpl
smerlos: Left as-is for now: the playbooks run `serial: 1`, so it's already one node at a time — an outer `--limit` loop would just double that up. On-prem needs the loop because its playbooks aren't serialized. Matching it here also pulls in the playbook prefix rename + `--upgrade-node`, so I opened #689 for it.

internal/apis/kfd/v1alpha2/immutable/create/infrastructure.go
smerlos: Good point on the coupling. Can't just drop it though — the infra roles (containerd, haproxy) read those vars and have no defaults. So I moved it: infra now has its own `group_vars/all.yml.tpl` with just the 3 vars it needs, rendered by the same walk the kubernetes phase uses. No more reaching into the `kubernetes/` folder, and `renderVersionVarsFile` is gone — net smaller diff.

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
smerlos: Kept it. Gives a clear "manifest has no versions at all" vs "this version isn't listed" — different fixes for whoever hits it.

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
smerlos: The vars are still needed (containerd/haproxy above), but you're right the wiring was the problem. With the move, infra no longer reaches into the `kubernetes/` folder — it renders its own template via the standard walk — so the special-case renderer and the cross-phase path are gone.

smerlos:
Thanks for the review — addressed all of it:
- Moved the infra version vars to their own template (`templates/infrastructure/immutable/ansible/group_vars/all.yml.tpl`); infra no longer renders from the `kubernetes/` folder and `renderVersionVarsFile` is gone.
- Dropped the `v`-strip (version passed as-is) and the duplicate validate-time check (+ `immutable.go`).
- Removed the two echo-only distribution hook stubs; kept the `ErrNoKubernetesVersions` guard.
- `--limit`/on-prem parity is tracked in #689.

Tested by installing both a dedicated-etcd and a stacked cluster (infra `group_vars/all.yml` renders `containerd_sandbox_image` on both). Paired distribution change: sighupio/distribution#544.

configs/upgrades/immutable/1.34.8-1.35.5/pre-kubernetes.sh.tpl
ralgozino: I don't see how `serial:1` and `--limit` are incompatible, you can limit with a group name too, for example. If you pass `--limit <single node>` it will run just fine even with `serial: 1`.

Having said that, the single node upgrade was already cabled and is working:
https://github.com/sighupio/furyctl/blob/6b54944283073060290c47b0217b54d6b6f939d7/internal/apis/kfd/v1alpha2/immutable/create/kubernetes.go#L56-L62

ralgozino:
please check the linting errors triggered by the new commits

configs/upgrades/immutable/1.34.8-1.35.5/pre-kubernetes.sh.tpl
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env sh
+
+set -eu
+
+kubectlbin="{{ .paths.kubectl }}"
+
+{{- if index .spec "kubernetes" }}
```
ralgozino: what is this `if` for?

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
```diff
@@ -223,43 +229,79 @@ func (i *Infrastructure) getSSHPublicKeyContent() (string, error) {
...
-// This manifest contains all versioning information for the Immutable installer.
-func (*Infrastructure) parseImmutableInstallerSpec(path string) (*immutableManifest, error) {
-	data, err := os.ReadFile(path)
+// selectImmutableAssets loads the vendored immutable.yaml and selects the block for the given
+// kubernetes version. It is the single selector used by every furyctl phase (infrastructure and
+// kubernetes), so baked artifacts and rendered role variables never disagree on version. A free
+// function (no Infrastructure receiver) so both phases call it directly. The version is passed as-is
+// (it comes from the kfd manifest, which we control), so a mismatch surfaces as
+// ErrKubernetesVersionNotFound; an empty manifest surfaces as ErrNoKubernetesVersions.
+func selectImmutableAssets(phasePath, kubeVersion string) (assets, error) {
```
ralgozino: this func is now called from several phases, it may be a good idea to move it to a different file inside the create package instead of having it here the infrastructure_bootstrap file

configs/upgrades/immutable/1.34.8-1.35.5/pre-kubernetes.sh.tpl
smerlos: You're right on both — thanks, and sorry for the confusion. `--limit` and `serial: 1` are orthogonal (host selection vs batching), so they compose fine; my "double serialization" wording was wrong. And I'd missed that single-node is already wired via `--upgrade-node` → `Playbook("upgrade-worker-nodes.yml", "--limit", k.upgradeNode)` (kubernetes.go:56-62). So there's nothing to change here: full-cluster uses the playbooks' `serial: 1`, single-node is `--upgrade-node`. I'll close the follow-up issue I'd opened (#689) since the capability already exists.

configs/upgrades/immutable/1.34.2-1.35.0/pre-kubernetes.sh.tpl
ralgozino, line 1: note, that upgrade path are for SD versions and not kubernetes, i.e. it should be 1.34.2 -> 1.35.0 and not 1.34.8->1.35.5

configs/upgrades/immutable/1.34.8-1.35.5/pre-kubernetes.sh.tpl
smerlos: This was a copy-paste leftover from onprem. Immutable does not use it. 
Action: removed

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
smerlos: Moved it to immutable_assets.go in the same create package — it's called from both phases.

configs/upgrades/immutable/1.34.2-1.35.0/pre-kubernetes.sh.tpl
smerlos: Renamed to `1.34.2-1.35.0` (SD versions), matching onprem.

ralgozino (approved):
LGTM 🏁 🚢 

well done!

PR #694, opened 2026-07-07
feat: show download progress

ralgozino:
### Summary 💡

Show download progress for assets > 5MB.

### Description 📝

Implement download progress tracking and content length probing

Show the user the progress of downloads for assets greater than 5MB so it the application does not seem unresponsive on slow connections or big files being downloaded (like flatcar assets).

The progress is not updated dynamically when no-tty is set.

Note that this applies only to HTTP/s3 downloads, for downloads using git (like the distribution or the module's repositories) this is not possible. 

Video (download speed is a little slow so it takes a lot of time):
https://github.com/user-attachments/assets/de1932bb-f165-407e-a520-a6ffda1b28cd

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] Tested that downloads progress is correctly shown with a clean .furyctl folder 
- [x] Tested that logs are not dynamic when --no-tty flag is set.

### Future work 🔧

None

> [!IMPORTANT]
>

~~- [ ] I've updated the `docs/releases/unreleased.md` file (or equivalent)~~ N.A.
- [x] CI is green


pkg/x/net/progress.go
```diff
@@ -0,0 +1,163 @@
...
+
+		return
+	}
+
+	// Rewrite the line in place: \r returns to column 0, then \033[2K clears it.
+	_, _ = fmt.Fprintf(r.tracker.out, "\r\033[2K  Downloading %s: %s", r.name, status)
+}
+
+// humanizeBytes formats a byte count using metric (base-1000) units.
+func humanizeBytes(b int64) string {
```
marcopaggioro, line 152: Have you considered using libraries like https://github.com/dustin/go-humanize/ ?

pkg/x/net/progress.go
```diff
@@ -0,0 +1,163 @@
...
+// are hundreds of MB) aren't mistaken for a hung process.
+type downloadProgressTracker struct {
+	out io.Writer
+	// Draw an in-place animated line; otherwise log so CI/--no-tty stay readable.
+	tty bool
+	// Size from a HEAD probe, used when the GET response has no Content-Length.
+	fallbackTotal int64
+}
+
+// newProgressTracker is a package var so tests can inject a tracker writing to a buffer.
```
marcopaggioro: It seems to me that this function is used not only by tests. Perhaps the comment is obsolete?

ralgozino: nice catch, thanks

pkg/x/net/progress.go
ralgozino: it seems a little overkill to me to introduce an external dependency for something like this (we are not using the library elsewere AFAIK). I would use something from the std-lib if it exists though

pkg/x/net/progress.go
ralgozino: comment updated

smerlos (approved):
LGTM

marcopaggioro (approved):

PR #697, opened 2026-07-09
fix: check that openvpn binary is present even if vpn-auto-connect argument is not present

marcopaggioro:
### Summary 💡

The openvpn binary is required to create VPN VM. But the presence of its binary is checked only if the --vpn-auto-connect flag is present.
If the binary is missing, the installation process will freeze indefinitely rather than failin.

### Description 📝

To prevent the installation from hanging, the installer now strictly validates the presence of the OpenVPN binary during the pre-flight checks, regardless of whether the `--vpn-auto-connect` argument is present.

Plus point: added units tests regarding the use of the `--vpn-auto-connect` flag.

---

### Before the fix
```
# spec.infrastructure.vpn configured and no --vpn-auto-connect flag
furyctl create cluster --outdir $PWD

{"level":"debug","action":"tofu apply -no-color -json plan/terraform.plan","msg":"{\"@level\":\"info\",\"@message\":\"module.vpn[0].null_resource.init: (local-exec): 2026-07-09 16:33:56.830115 I | openvpn.go:102: exec: \\\"openvpn\\\": executable file not found in $PATH\",\"@module\":\"tofu.ui\",\"@timestamp\":\"2026-07-09T16:33:56.830203+02:00\",\"hook\":{\"resource\":{\"addr\":\"module.vpn[0].null_resource.init\",\"module\":\"module.vpn[0]\",\"resource\":\"null_resource.init\",\"implied_provider\":\"null\",\"resource_type\":\"null_resource\",\"resource_name\":\"init\",\"resource_key\":null},\"provisioner\":\"local-exec\",\"output\":\"2026-07-09 16:33:56.830115 I | openvpn.go:102: exec: \\\"openvpn\\\": executable file not found in $PATH\"},\"type\":\"provision_progress\"}\n","time":"2026-07-09T16:30:13+02:00"}
[...]

PR #700, opened 2026-07-15
fix: don't report openvpn as found when OpenVPN binary is not checked

marcopaggioro:
### Summary 💡

`furyctl validated dependencies` is returning misleading output regarding the presence of the OpenVPN binary.
This PR aims to remove that log.

### Description 📝

In #697 we fixed the check for OpenVPN binary from the command `furyctl validated dependencies`.
The OpenVPN binary presence is not checked if not necessary.
The side-effect of that fix is that even if the check is not done, the command output is contradictory because it says that the binary is found but in reality the check is not done at all:
```
...
INFO furyagent: binary found in vendor folder     
INFO openvpn: binary found in vendor folder 
...
```

After this fix:
```
...
INFO furyagent: binary found in vendor folder  
...
```

### Breaking Changes 💔

Not detected.

### Tests performed 🧪

Tested locally that the

### Future work 🔧

Not planned.

- [ ] ~I've updated the `docs/releases/unreleased.md` file (or equivalent)~
- [x] CI is green

### DIFF
--- a/internal/apis/kfd/v1alpha2/ekscluster/tool.go
+++ b/internal/apis/kfd/v1alpha2/ekscluster/tool.go
@@ -26,17 +26,27 @@ func NewExtraToolsValidator(executor execx.Executor) *ExtraToolsValidator {
 }
 
 func (x *ExtraToolsValidator) Validate(confPath string) ([]string, []error) {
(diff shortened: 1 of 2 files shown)


ralgozino (approved):
lgtm

PR #725, opened 2026-07-23
feat(immutable): validate node roles

ralgozino:
### Summary 💡

Validate that:
- Nodes have one and only one role defined in the furyctl.yaml file.
- There are no references to nodes not managed by furyctl.

### Description 📝

Validate (during the validate config step) that the configuration file for an Immutable cluster that:
- All nodes in `.spec.infrastructre.nodes` have _one and only one_ role assigned (i.e. they are referenced as members of load balancers, the control plane, etcd, or node groups).
- All referenced nodes/members in load balancers, the control plane, etcd, and node groups have the corresponding node definition in `.spec.infrastructre.nodes`; i.e. all nodes are managed by furyctl.

With this validation, there's no more risk of defining a node in `.spec.infrastructre.nodes` and then forgetting to use it afterward as member of some role.

Also, we had a fallback to the worker role when the node was not in the other categories. This is not possible anymore, a node that has no role will trigger an error instead of silently being used as a worker and generating all (wrong and unused) the butane/ignition configuration for it.

New errors:
- **`validateNodeRoles`** — every node in `.spec.infrastructure.nodes` is
  referenced by at least one role list; otherwise it errors listing the orphan
  hostnames.
- **`validateNodeReferences`** — every hostname referenced by a role list has a
  matching `.spec.infrastructure.nodes` entry; dangling references are reported.
- **`validateSingleRole`** — a hostname must not appear in more than one role
  list, e.g. `ctrl01 (controlplane, etcd)`. Stacked etcd is expressed by
  **omitting** the `.spec.kubernetes.etcd` block (etcd co-located on the control
  plane); an explicit `etcd` block means dedicated etcd nodes and must not repeat
  control-plane hostnames.

Supporting changes:

- **Single source of truth for role derivation** in the curated public config
  (`public.ImmutableKfdV1Alpha2`): `RoleAssignments()` enumerates every
  `(hostname, role)` pair in role order, and `NodeRole()` / the validators derive
  from it. Added the `NodeRoleNone/ControlPlane/LoadBalancer/Etcd/Worker`
  constants.
- **Schema**: added `SpecKubernetes.NodeGroups` (`SpecKubernetesNodeGroup`),
  mirroring the distribution's `immutable-kfd-v1alpha2.json`. The four role-list
  paths were verified against that schema (they are exactly its four
  `Spec.Kubernetes.Member` references).

Updated some test cases that had outdated structure.

### Breaking Changes 💔

None, this is the right behaviour, any other configuration should have resulted in an error at runtime (e.g. nodes not being used, configured, etc.) that should have been detected by the operator.

### Tests performed 🧪

- [x] Successfuly created an Immutable cluster from scratch.
- [x] Tested adding a node without using it as a member of the roles produces the right error message via `furyctl validate config`.
- [x] Tested adding a node with more than one role produces the right error message via `furyctl validate config`.
- [x] Tested that referencing a hostname that is not declared as node produces the right error message via `furyctl validate config`.

### Future work 🔧

None

- [x] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change — scoped to Immutable only
- [x] CI is green

### DIFF
--- a/internal/apis/kfd/v1alpha2/immutable/schema.go
+++ b/internal/apis/kfd/v1alpha2/immutable/schema.go
@@ -4,8 +4,136 @@
 
 package immutable
 
+import (
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/sighupio/furyctl/internal/apis/kfd/v1alpha2/immutable/public"
+	yamlx "github.com/sighupio/furyctl/pkg/x/yaml"
+)
+
+// roleListPaths names the config paths that assign roles, shared by the messages below.
+const roleListPaths = ".spec.kubernetes.controlPlane.members, .spec.kubernetes.etcd.members, " +
+	".spec.kubernetes.nodeGroups[].nodes or .spec.infrastructure.loadBalancers.members"
+
+var (
+	ErrNodeWithoutRole = errors.New(
+		"every node defined in .spec.infrastructure.nodes must be assigned at least one role " +
+			"(referenced in " + roleListPaths + ")",
+	)
+	ErrNodeNotDefined = errors.New(
+		"every hostname referenced by a role (" + roleListPaths +
+			") must have a matching entry in .spec.infrastructure.nodes",
+	)
+	ErrNodeMultipleRoles = errors.New(
+		"a node must be assigned a single role, but these hostnames are referenced by more than one of " +
+			roleListPaths +
+			" (for stacked etcd omit the .spec.kubernetes.etcd block instead of repeating hostnames)",
+	)
+)
+
 type ExtraSchemaValidator struct{}
 
-func (*ExtraSchemaValidator) Validate(_ string) error {
+func (*ExtraSchemaValidator) Validate(confPath string) error {
+	conf, err := yamlx.FromFileV3[public.ImmutableKfdV1Alpha2](confPath)
+	if err != nil {
+		return err
+	}
+
+	// Cross-check node lists and role lists: every node has a role, every referenced
+	// hostname is a defined node, and no hostname holds more than one role. Report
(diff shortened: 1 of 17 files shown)


internal/apis/kfd/v1alpha2/immutable/schema.go
```diff
@@ -4,8 +4,136 @@
...
+		return fmt.Errorf("%w: %s", ErrNodeNotDefined, strings.Join(dangling, ", "))
+	}
+
+	return nil
+}
+
+// validateSingleRole asserts no hostname is referenced by more than one role list.
+// Roles are gathered per hostname (first-seen order) for a message like
+// "ctrl01 (controlplane, etcd)".
+func validateSingleRole(conf *public.ImmutableKfdV1Alpha2) error {
```
stefanoghinelli: Should the same hostname be allowed in multiple spec.kubernetes.nodeGroups[].nodes entries?
In `validateSingleRole`, all node groups are grouped into a single worker role; therefore, a node that appears in two node groups would seem to pass validation.

But I might be missing some logic :)

stefanoghinelli (commented):
One question, lgtm otherwise

internal/apis/kfd/v1alpha2/immutable/schema.go
ralgozino: That's a good question, I guess you should not be able to use the same node in two different node groups.

I will fix it, thanks

ralgozino: fixed

stefanoghinelli (approved):

smerlos (approved):
LGTM

PR #727, opened 2026-07-24
chore: continuos quality improvements

marcopaggioro:
### Summary 💡

Continuos Quality Improvements

### Description 📝

This PR aims to reduce boilerplate code wthout changing code logic.

### Breaking Changes 💔

Not detected.

### Tests performed 🧪

CI: 

### Future work 🔧

Not planned.

> [!IMPORTANT]
>

- [ ] ~I've updated the `docs/releases/unreleased.md` file (or equivalent)~
- [x] CI is green


cmd/serve/status_table.go
```diff
@@ -59,21 +59,11 @@ type nodeStatusTable struct {
...
-
-	slices.Sort(order)
-
 	return &nodeStatusTable{
 		out:       f,
 		tty:       execx.ShouldAnimate(f),
-		order:     order,
-		status:    status,
+		order:     slices.Sorted(maps.Keys(initial)),
+		status:    maps.Clone(initial),
```
ralgozino: as FYI from Claude:
> `maps.Clone(nil)` returns `nil`, whereas the old `make(map[string]string, len(initial))` was always non-nil. `Update()` (line 93) writes `t.status[node] = status`, which would panic on a nil map.
>
> Both current callers seed a non-nil map (`make(...)` in infrastructure.go:158 and `map[string]string{}` in cmd/serve.go:48), and `maps.Clone` of a non-nil empty map stays non-nil — so no active bug. But the defensiveness the old code had for free is now gone.

cmd/apply.go
```diff
@@ -572,15 +572,15 @@ func setupApplyCmdFlags(cmd *cobra.Command) {
 
 	// Tab-autocomplete for post-apply-phases.
 	if err := cmd.RegisterFlagCompletionFunc("post-apply-phases", func(_ *cobra.Command, _ []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		// The post-apply-phases flag accepts a comma separated list of phases, so we need to take the passed list and add a new valid option at the end of it.
+		// The post-apply-phases flag accepts a comma-separated list of phases, so we need to take the passed list and add a new valid option at the end of it.
 		phases := cluster.MainPhases()
 		toCompleteList := strings.Split(toComplete, ",")
 		toCompleteLast := toCompleteList[len(toCompleteList)-1]
-		completion := []string{}
+		var completion []string
```
ralgozino, line 579: I'm curious, why is this better? is this more idiomatic?

internal/apis/config/validation.go
```diff
@@ -57,32 +58,22 @@ var (
...
+		"aws-region":            ValidateAwsRegion,
 	}
 
-	if err = validate.RegisterValidation("aws-region", ValidateAwsRegion); err != nil {
-		return nil
+	for name, fn := range validations {
+		if err := validate.RegisterValidation(name, fn); err != nil {
+			logrus.Warnf("error registering validation %q: %v", name, err)
+			return nil
+		}
```
ralgozino:
```suggestion
	for name, validation := range validations {
		if err := validate.RegisterValidation(name, validation); err != nil {
			logrus.Warnf("error registering validation %q: %v", name, err)
			return nil
		}
```

internal/apis/kfd/v1alpha2/common/create/preupgrade.go
ralgozino, line 1: I don't see the benefits of the changes to this file. Is it more idiomatic this way?

internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
ralgozino, line 1: Probably an skill issue, but the old code was much more readable to me

internal/apis/kfd/v1alpha2/immutable/create/preflight.go
ralgozino, line 1: same as for eks, old code was more readable to my eyes

internal/apis/kfd/v1alpha2/kfddistribution/create/preflight.go
```diff
@@ -236,18 +237,11 @@ func (p *PreFlight) CheckStateDiffs(d r3diff.Changelog, diffChecker diffs.Checke
 		return nil
 	}
 
-	// Get all immutable rules.
-	immutableRules := r.GetImmutableRules("distribution")
-
-	// Filter out the rules that have matching safe conditions.
-	filteredRules := r.FilterSafeImmutableRules(immutableRules, d)
-
-	// Extract the paths from the filtered rules.
```
ralgozino, line 245: this change feels like code golfing to me, is there any benefit besides reducing the number of lines that I am missing?

internal/apis/kfd/v1alpha2/onpremises/create/preflight.go
ralgozino, line 1: same as above

internal/tool/runner.go
```diff
@@ -67,6 +67,8 @@ func NewRunnerFactory(executor execx.Executor, paths RunnerFactoryPaths) *Runner
 }
 
 func (rf *RunnerFactory) Create(name Name, version, workDir string) Runner {
+	binPath := filepath.Join(rf.paths.Bin, string(name), version, string(name))
```
ralgozino: calling this var `binPath` could be confusing with the `binPath` flag. I'd call it something different like `toolBinPath`

internal/x/exec/cmd.go
```diff
@@ -133,18 +133,11 @@ func (c *Cmd) Stop() error {
...
 
 	defer cancel()
 
-	if len(c.Args) == 1 {
-		cmdCtx = exec.CommandContext(ctx, c.Path)
-	} else {
-		args := c.Args[1:]
-		cmdCtx = exec.CommandContext(ctx, c.Path, args...)
-	}
+	cmdCtx := exec.CommandContext(ctx, c.Path, c.Args[1:]...)
```
ralgozino, line 140: TIL, nice

pkg/template/config.go
```diff
@@ -110,12 +110,6 @@ func NewConfigWithoutData(tplSource *merge.Merger, excluded []string) (Config, e
 }
 
 func newTemplatesFromMap(t any) (*Templates, error) {
-	var exc []string
-
-	var inc []string
-
-	var err error
-
```
ralgozino, line 118: in the other files you prefered to change from declarations with assignment (e.g. `s := "sample"` to declarations and then assign values (e.g. `var s string and then s = "sample"`; but here you are doing the opposite.

I'm curious why

ralgozino (changes requested):
overall these are very good changes! I think it is a nice improvement to the code base.

I left some comments, there's nothing actually blocking but in some places I think the improvements may have been a little too much and ended up with less readability, but I don't discard a skill issue on my side :)

internal/tool/runner.go
marcopaggioro: Renamed!

cmd/apply.go
marcopaggioro: The old method initializes the slice, while the new one doesn't. 
Essentially, the old approach created an empty slice immediatly overwritten by line 584, so an unused slice. The new approach doesn't create an empty slice.

cmd/serve/status_table.go
marcopaggioro: Aligned the code, TY

internal/apis/kfd/v1alpha2/common/create/preupgrade.go
marcopaggioro: This is safer code.
In the assignment `p.upgrade.From, ok = distributionVersionChange.From.(string)`, the value of `p.upgrade.From` is overwritten even if `ok` is false.
By using a temporary variable, we avoid this risk.

pkg/template/config.go
marcopaggioro: Uhm I'm not sure I understand correctly.
I never changed from assignment to declaration+assignment (from `s := "sample"` to `var s string and then s = "sample"`).

The rule I applied is:
- Where the value was overwritten before any read, I dropped the useless initialization and kept only the declaration: `distrodl := &dist.Downloader{}` -> `var distrodl *dist.Downloader`.
- Where a variable is only appended in a loop, initialization is useless but the declaration before the loop is required: `args := []string{}` -> `var args []string`.
- Where the declaration itself was unnecessary, like in this method, where `var inc []string` and `var err error` existed only to allow `=` instead of `:=` -> I removed it and used `:=` at the point of use.

internal/apis/kfd/v1alpha2/kfddistribution/create/preflight.go
marcopaggioro: Not exactly code golf here :)
I actually preferred using an existing method, `slicesx.Map` (it creates a new slices and applies a function to each value in the map), which does exactly what this method did.

Plus point: `slicesx.Map` is also more efficient: it pre-allocates the result slice with exact capacity (`make([]U, len(s))`) and fills it by index, while the old loop started from a zero-cap slice and grew it via append, paying for the reallocations and copies on each growth step (1, 2, 4, 8...).

BTW I did a bit of refactor to increase redability

marcopaggioro:
@ralgozino I did a bit of refactor to the three files where you reported poor readability. 
Try taking a look now and let me know :)

cmd/apply.go
ralgozino: AFAIK (and I just checked) both allocate memory (one is nil the other is empty), but being that we are appending in 584 they is no practical difference. It is more idiomatic to use declare without assignment though.

internal/apis/kfd/v1alpha2/common/create/preupgrade.go
ralgozino: make sense, thanks

ralgozino (approved):
looks good to me, thanks for the detailed answers :)

PR #728, opened 2026-07-24
feat(immutable): download python sysext with checksum verification

smerlos:
### Summary 💡

Sysext packages are downloaded by furyctl and staged on the boot server, but nothing checked that what
landed on disk is what the manifest pins. This verifies every sysext package that declares a `sha256`
in `immutable.yaml` right after download, failing the phase on mismatch. Packages without a pin keep
the previous behaviour, so nothing changes for the manifests in use today.

Motivation is air-gap safety: the Python sysext is moving off the public Flatcar release server onto
our own infrastructure, and an artifact we serve ourselves should be checked against a digest that
travels with the manifest instead of being trusted blindly.

Closes:

Relates: sighupio/distribution#571, sighupio/installer-immutable#18, sighupio/distribution#567

### Description 📝

Two files under `internal/apis/kfd/v1alpha2/immutable/create/`:

- `sysextArchInfo` gains an optional `sha256` field, matching the key used by the manifest entries in
  sighupio/installer-immutable#18. Decoding is a non-strict `yaml.Unmarshal`, so old and new manifests
  work against old and new furyctl.
- New `verifySHA256` helper: streams the file through `sha256.New()` with `io.Copy` (a `.raw` is
  hundreds of MB, never held in memory) and compares hex with `strings.EqualFold`. On mismatch it
  returns a wrapped `ErrSysextChecksumMismatch` carrying wanted and actual digests, matchable with
  `errors.Is`.
- `downloadSysextPackages` verifies right after `DownloadWithMode`. No pin → logs that verification
  was skipped and continues; pin that does not match → aborts the phase.
- The per-arch `sha256` is also carried into `sysext_targets` next to the existing `url`.
- Unit test covering matching and mismatching digests.

### Breaking Changes 💔

None. The field is optional in both directions: furyctl with this change against a manifest without
`sha256` behaves as before; furyctl without it against a manifest with `sha256` ignores the unknown
key.

### Tests performed 🧪

- [x] `mise run test-unit`, `mise run lint` (0 issues), `mise run license-check`,
      `mise run test-integration`, `mise run test-flags`, `go build ./...` — all pass on the rebased branch
- [x] Live install on a 6-node Immutable cluster (k8s 1.34.8 / Flatcar 4593.2.1): the pinned
      `flatcar-python` was verified (`Verified flatcar-python-4593.2.1-x86-64.raw against pinned sha256`),
      the on-disk sha256 matches the manifest pin, and the 5 unpinned packages downloaded unchanged
- [x] All 8 nodes fetched the sysext from the local PXE asset server (8/8 HTTP 200), 0 requests to the
      public Flatcar release server; `flatcar-python` merged on the nodes and `python3` available
- [x] Live upgrade path (Flatcar 4593.2.1 → 4593.2.3)
- [ ] CI is green

### Future work 🔧

The Ansible `sysext` role still verifies node-side downloads against a `SHA256SUMS` file served from
the same host as the artifact. Consuming the per-arch `sha256` the manifest already pins would be
stronger, since the digest would no longer come from the host being verified.

- [x] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change — only Immutable is affected
- [ ] CI is green

### DIFF
--- a/internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
+++ b/internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
@@ -852,7 +852,7 @@ func (*Infrastructure) downloadSysextPackages(
 
 			// Only packages that pin a sha256 are verified; the rest keep the previous behaviour.
 			if archInfo.SHA256 == "" {
-				logrus.Warnf("Sysext package %s (%s) has no pinned sha256, skipping verification", pkg.Name, arch)
+				logrus.Debugf("Sysext package %s (%s) has no pinned sha256, skipping verification", pkg.Name, arch)
 
 				continue
 			}
--- a/internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
+++ b/internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
@@ -7,9 +7,12 @@ package create
 import (
 	"bytes"
 	"compress/gzip"
+	"crypto/sha256"
 	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -34,6 +37,7 @@ var (
 	ErrButaneConversionFatal     = errors.New("butane conversion fatal errors")
 	ErrButaneFatalErrors         = errors.New("butane translation has fatal errors")
 	ErrImmutableConfigMalformed  = errors.New("immutable furyctl config is malformed")
+	ErrSysextChecksumMismatch    = errors.New("sysext package checksum mismatch")
 )
 
 // defaultNodeArch mirrors the schema default of spec.infrastructure.nodes[].arch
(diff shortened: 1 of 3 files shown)


ralgozino (changes requested):
PR description does not follow our template and CI is green on main, please rebase or check your changes

internal/apis/kfd/v1alpha2/immutable/create/infrastructure_bootsrap.go
```diff
@@ -821,6 +849,19 @@ func (*Infrastructure) downloadSysextPackages(
 			); err != nil {
 				return fmt.Errorf("error downloading %s for %s: %w", pkg.Name, arch, err)
 			}
+
+			// Only packages that pin a sha256 are verified; the rest keep the previous behaviour.
+			if archInfo.SHA256 == "" {
+				logrus.Warnf("Sysext package %s (%s) has no pinned sha256, skipping verification", pkg.Name, arch)
```
ralgozino: I would not log this in `WARN`, the user can't do anything about it and WARN messages are supposed to be actionable.

This is internal information that should be in DEBUG in my opinion:

```suggestion
				logrus.Debugf("Sysext package %s (%s) has no pinned sha256, skipping verification", pkg.Name, arch)
```

ralgozino (changes requested):

ralgozino (approved):

PR #739, opened 2026-07-29
fix(on-prem): don't treat any preflight failure as "cluster doesn't exist"

### LINKED ISSUE #717: on-prem: any preflight verify-playbook.yaml failure is treated as "cluster does not exist" and skip the upgrade diff check
### Describe the bug

In the OnPremises provider the `PreFlight.Exec()` treats any failure of `verify-playbook.yaml` as proof that the cluster does not exist yet, regardless of why the playbook failed:

https://github.com/sighupio/furyctl/blob/c2349e8b9788069a152bf3b2cf51a7f05bb793e5/internal/apis/kfd/v1alpha2/onpremises/create/preflight.go#L143-L151

`verify-playbook.yaml` itself only fails on purpose when no master has /etc/kubernetes/admin.conf. But ansible-playbook exits non-zero on any failed task on any host and preflight.go does not distinguish that specific case from any other failure (checksum mismatch, timeout, etc). 

As recently happened the "Getting admin.conf kubeconfig" task fetches every master's admin.conf to the same local path in parallel (flat: yes, no run_once), so it races and fails with checksum mismatch.

When this happens on an existing, healthy cluster, furyctl skips the config diff entirely, so the distributionVersion change is never detected, the upgrade warning and confirmation prompt never show even with --upgrade, and furyctl runs `create-playbook.yaml` instead of kubeadm upgrade node.

### To Reproduce

1. On an existing SD on-prem cluster with 3 control-plane nodes, cd into <outdir>/.furyctl/<cluster-name>/preflight/.
2. Run verify-playbook.yaml in a loop:
```bash
   for i in $(seq 50); do ansible-playbook verify-playbook.yaml | grep -q "checksum mismatch" && echo "hit at $i" && break; done
```
3. Within a few iterations the "Getting admin.conf kubeconfig" task fails with a checksum mismatch, because every master fetches to the same local destination.
4. After that, change distributionVersion in `furyctl.yaml` and run `furyctl apply --upgrade`. It will be executed `create-playbook.yaml` instead of the upgrade playbooks.

### Expected behavior

`PreFlight.Exec()` should only take the "cluster does not exist" shortcut when `verify-playbook.yaml` failed for the specific, intended reason (no master has admin.conf anywhere). Any other failure (checksum mismatch, permissions, etc.) should abort apply with a clear, explicit error, not fall through silently to a plain reconciliation run on what is actually an existing cluster.

### Logs 

```
TASK [Getting admin.conf kubeconfig] *******************************************
fatal: [furypass-controlplane-2]: FAILED! =>
changed: false
checksum: 3acc7952c0281ea1ef49b6fb4e1ec9f239c4891b
dest: .../preflight/admin.conf
file: /etc/kubernetes/admin.conf
md5sum: fbcfeeb09e378a33627b02b80dbff865
msg: checksum mismatch
remote_checksum: 361e6d771e8c11fb1e078f244c44f80b2ad9ffa7
remote_md5sum: null
changed: [furypass-controlplane-0] =>
changed: true
checksum: 3acc7952c0281ea1ef49b6fb4e1ec9f239c4891b
remote_checksum: 3acc7952c0281ea1ef49b6fb4e1ec9f239c4891b
changed: [furypass-controlplane-1] =>
changed: true
checksum: d9500ba789dd394459d31c35fdbb95d63be9c510
remote_checksum: d9500ba789dd394459d31c35fdbb95d63be9c510

PLAY RECAP ***********************************************************************
furypass-controlplane-0    : ok=7  changed=1  unreachable=0  failed=0  skipped=1
furypass-controlplane-1    : ok=7  changed=1  unreachable=0  failed=0  skipped=1
furypass-controlplane-2    : ok=2  changed=0  unreachable=0  failed=1  skipped=0
```

### Software versions

This behavior has been present since v0.27.0 (the first release that include on-premises kind)

### Additional context

Related to: [distribution#561](https://github.com/sighupio/distribution/pull/561)

stefanoghinelli:
### Summary 💡

Makes furyctl distinguish a "the cluster doesn't exist yet" from any other preflight failure on OnPremises clusters, so a flaky host or a local fetch error no longer gets silently treated as a fresh install and skips upgrade and immutability checks.

Closes:
- #717

Relates:
- [distribution#577](https://github.com/sighupio/distribution/pull/577)

### Description 📝

Before this change, `PreFlight.Exec()` treated any error from `verify-playbook.yaml` as proof the cluster was brand new. That playbook used a `fail:` task as its only way to say "no master has admin.conf yet", and furyctl had no way to tell that apart from a host being unreachable, a checksum mismatch during the kubeconfig fetch, or any other unrelated Ansible error.

This PR splits the check into two independent steps:
- If `verify-playbook.yaml` itself fails, that's now always treated as a real error and propagated.
- Cluster existence is decided only after a successful Ansible run, by checking whether `admin.conf` actually landed in the local working directory.

### Breaking Changes 💔

None.

### Tests performed 🧪

All performed manually against an on-prem cluster (3 control planes, 1 infra node)

- [x] Tested a cluster creation. Still correctly recognized as "cluster doesn't exist".
- [x] Tested a plain `apply` on an existing cluster.
- [x] Tested a real `--upgrade` with a `distributionVersion` change. Diff still correctly detected, real kubeadm upgrade path runs.
- [x] Tested a forced local failure (fetch destination not reachable). Now fails with an explicit error instead of "cluster doesn't exist".
- [x] Tested a forced Ansible failure. Fails with an explicit error.
- [x] Tested adding a 4th control-plane node in `masters.hosts`, simulating an immutable edit. Cluster still correctly recognized as existing and the immutability check on `masters.hosts.*` correctly blocked the change.

### Future work 🔧

Apply the same fix to the Immutable kind

> [!IMPORTANT]
>

- [x] CI is green


ralgozino (approved):
lgtm

PR #741, opened 2026-08-05
feat(pki): check the PKI folder before an apply

### LINKED ISSUE #584: immutable: pki folder not validated
### Describe the bug

furyctl apply does not validate the presence of the `pki` folder, resulting in an ansible error at runtime instead of gracefully erroring out.

### To Reproduce
1. Create an Immutable cluster configruation file
2. **Don't** create the pki folder
3. Run `furyctl apply`
4. See that the infrastructure phase passes but the kubernetes phase fails with an ansible error like below:

```
<<< caused by >>>

Could not find or access '<redacted>/pki/etcd/ca.key' on the Ansible Controller. If you are using a module and expect the file to exist on the remote, see the remote_src option.

failed: [cp1.flatcar] (item=ca.key) =>
    ansible_loop_var: item
    changed: false
    item: ca.key
    msg: 'Unexpected AnsibleActionFail error: Could not find or access ''<redacted>/pki/etcd/ca.key''
        on the Ansible Controller.'
failed: [cp2.flatcar] (item=ca.key) =>
    ansible_loop_var: item
    changed: false
    item: ca.key
    msg: 'Unexpected AnsibleActionFail error: Could not find or access ''<redacted>/pki/etcd/ca.key''
        on the Ansible Controller.'
failed: [cp3.flatcar] (item=ca.key) =>
    ansible_loop_var: item
    changed: false
    item: ca.key
    msg: 'Unexpected AnsibleActionFail error: Could not find or access ''<redacted>/pki/etcd/ca.key''
        on the Ansible Controller.'

PLAY RECAP *********************************************************************
cp1.flatcar                : ok=18   changed=0    unreachable=0    failed=1    skipped=7    rescued=0    ignored=0
cp2.flatcar                : ok=18   changed=0    unreachable=0    failed=1    skipped=7    rescued=0    ignored=0
cp3.flatcar                : ok=18   changed=0    unreachable=0    failed=1    skipped=7    rescued=0    ignored=0


err:
```

### Expected behavior

furyctl should validate the presence of the needed pki files and gracefully handle when they are not present

### Software versions
Please complete the following information:

 - SIGHUP Distribution version: v1.35.1
 - furyctl version: 0.35.1

### Additional context

Haven't verified yet but the same issue may affect `OnPremises`

ralgozino:
### Summary 💡

furyctl did not check the folder in `.spec.kubernetes.pkiPath` (Immutable) or `.spec.kubernetes.pkiFolder` (OnPremises) before an apply.

When the folder or one of its files was absent, the apply failed in the middle with an Ansible error hard to understand.

Closes https://github.com/sighupio/distribution/issues/584

Relates:
- https://github.com/sighupio/distribution/pull/585
- https://github.com/sighupio/distribution/pull/583

### Description 📝

furyctl did not check the folder in `.spec.kubernetes.pkiPath` (Immutable) or `.spec.kubernetes.pkiFolder` (OnPremises) before an apply. When the folder or one of its files was absent, the apply failed in the middle, inside an Ansible task. The message did not say how to correct the fault.

`furyctl apply` and `furyctl validate config` now check that the folder holds the files that the installer roles copy to the nodes:

- under master/: ca.crt, ca.key, front-proxy-ca.crt, front-proxy-ca.key, sa.key and sa.pub
- under etcd/: ca.crt and ca.key

`furyctl apply` skips the check when the selected phases don't need the PKI. As a result, an apply of the distribution or the plugins phase alone does not ask for the CA keys. A dry run also skips the check, with a warning, because it stops before Ansible.

#### Failure modes

An absent folder names the `furyctl create pki` command to run.

An incomplete folder, names the files that are absent or empty.

A file that furyctl cannot read gives a read error, and not a name in that list. The message for an incomplete folder tells the user to delete the folder, and that step destroys the CA of a cluster that exists.

The table below gives the commands and the phases that do the check:

| Command | The check |
|---|---|
| `apply`, `apply --phase infrastructure\|kubernetes` | runs |
| `apply --phase distribution\|plugins` | does not run |
| `apply --start-from` a distribution or plugins sub-phase | does not run |
| `apply --start-from post-kubernetes` | does not run, because `coreKubernetes` skips `apply.yaml` |
| `apply --post-apply-phases kubernetes` | runs, because `extraPhases` repeats that phase |
| `apply --dry-run` | does not run, and gives a warning |
| `validate config` | runs |
| the other 6 commands that call `config.Validate` | do not run it |

Worth noting that during the resume of an upgrade the PKI check will be done because the start phase comes from the saved state in the configmap and is not known at validate config time. If the user specifies `--upgrade` and `--start-from` the starting phase will be known and the check will be skipped if not needed.

### Breaking Changes 💔

None. Each new check is preventive. The blocked cases make the apply fail later, inside an Ansible task.

A test in CI can fail if it does not create the PKI folder before it uses it. The correction is simple: create the pki first. From what I checked, distribution and furyctl tests are not affected.

### Tests performed 🧪

- [x] `furyctl create config --kind Immutable` and `--kind OnPremises`, against a local distribution: the file is written, and it is not deleted.
- [x] `validate config` with the folder absent, with a complete folder, and with 2 files removed.
- [x] `validate config` with `pkiPath: pki` (a relative value without a prefix), with `{path://realpki}` and with `{env://PKI_DIR}`.
- [x] `validate config` with a relative `--config` value from 2 different working folders: the message gives the same absolute path.
- [x] `validate config` with an Immutable configuration without `pkiPath`, against the schema of distribution#585: the schema reports the absent field first.
- [x] `apply --phase distribution` with an incomplete folder: the check does not run.
- [x] `apply --phase kubernetes` with an incomplete folder: exit code 1, with the message for an incomplete folder.
- [x] `apply --dry-run` with the folder absent: a warning, and the apply continues.
- [x] A folder that furyctl cannot read (mode `0o000`): a read error, and not a message for absent files.
- [x] `furyctl dump template` in `distribution/tests/templates/immutable/01-full-calico-nginx`, which holds `pkiPath: ./pki` and no `pki` folder: exit code 0, and the result is the same as the baseline. This is the command that `test:templates:regressions` runs.
- [x] An apply against a real Immutable cluster.

### Future work 🔧

- We could ask the user if wants to create the missing PKI folder instead of erroring out, but this feature needs some planning. Evaluate whether furyctl writes the path into `furyctl.yaml`.

> [!IMPORTANT]
>

- [x] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change — `NewPKIValidatorFactory` returns nil for `EKSCluster` and `KFDDistribution`, so their behavior does not change
- [x] CI is green

### DIFF
--- a/internal/apis/kfd/v1alpha2/immutable/public/schema.go
+++ b/internal/apis/kfd/v1alpha2/immutable/public/schema.go
@@ -95,6 +95,9 @@ type SpecKubernetes struct {
 	ControlPlane SpecKubernetesControlPlane `yaml:"controlPlane"`
 	Etcd         *SpecKubernetesEtcd        `yaml:"etcd,omitempty"`
 	NodeGroups   []SpecKubernetesNodeGroup  `yaml:"nodeGroups,omitempty"`
+	// PkiPath is the folder that holds the CA certificates and keys for the control plane and etcd. The
+	// playbooks read them from the `master` and `etcd` subfolders.
+	PkiPath *string `yaml:"pkiPath,omitempty"`
 }
 
 // SpecKubernetesNodeGroup assigns nodes the worker role by hostname.
--- a/internal/apis/schema.go
+++ b/internal/apis/schema.go
@@ -9,6 +9,7 @@ import (
 	"github.com/sighupio/furyctl/internal/apis/kfd/v1alpha2/immutable"
 	"github.com/sighupio/furyctl/internal/apis/kfd/v1alpha2/kfddistribution"
 	"github.com/sighupio/furyctl/internal/apis/kfd/v1alpha2/onpremises"
+	"github.com/sighupio/furyctl/internal/distribution"
 )
 
 type ExtraSchemaValidator interface {
@@ -26,12 +27,12 @@ type PKIValidator interface {
 // `furyctl create config` writes a configuration whose PKI folder does not exist yet.
 func NewPKIValidatorFactory(apiVersion, kind string) PKIValidator {
 	switch apiVersion {
-	case "kfd.sighup.io/v1alpha2":
+	case distribution.APIVersionV1Alpha2:
 		switch kind {
-		case "OnPremises":
+		case distribution.OnPremisesKind:
 			return &onpremises.PKIValidator{}
 
-		case "Immutable":
+		case distribution.ImmutableKind:
 			return &immutable.PKIValidator{}
 
 		default:
@@ -45,18 +46,18 @@ func NewPKIValidatorFactory(apiVersion, kind string) PKIValidator {
 
 func NewExtraSchemaValidatorFactory(apiVersion, kind string) ExtraSchemaValidator {
 	switch apiVersion {
-	case "kfd.sighup.io/v1alpha2":
+	case distribution.APIVersionV1Alpha2:
 		switch kind {
-		case "EKSCluster":
+		case distribution.EKSClusterKind:
 			return &ekscluster.ExtraSchemaValidator{}
 
-		case "KFDDistribution":
+		case distribution.KFDDistributionKind:
 			return &kfddistribution.ExtraSchemaValidator{}
 
-		case "OnPremises":
+		case distribution.OnPremisesKind:
 			return &onpremises.ExtraSchemaValidator{}
 
-		case "Immutable":
+		case distribution.ImmutableKind:
 			return &immutable.ExtraSchemaValidator{}
 
 		default:
--- a/internal/apis/schema.go
+++ b/internal/apis/schema.go
@@ -15,6 +15,34 @@ type ExtraSchemaValidator interface {
 	Validate(confPath string) error
 }
 
+// PKIValidator checks the local PKI folder that a configuration points at.
+type PKIValidator interface {
+	ValidatePKI(confPath string) error
+}
+
+// NewPKIValidatorFactory returns the PKI validator of the kind, or nil for a kind that reads no local
+// PKI. This check is separate from ExtraSchemaValidator because only `furyctl apply` and
+// `furyctl validate config` run it. The other commands must not run it. For example,
+// `furyctl create config` writes a configuration whose PKI folder does not exist yet.
+func NewPKIValidatorFactory(apiVersion, kind string) PKIValidator {
+	switch apiVersion {
+	case "kfd.sighup.io/v1alpha2":
+		switch kind {
+		case "OnPremises":
+			return &onpremises.PKIValidator{}
+
+		case "Immutable":
+			return &immutable.PKIValidator{}
+
+		default:
+			return nil
+		}
+
+	default:
+		return nil
+	}
+}
+
 func NewExtraSchemaValidatorFactory(apiVersion, kind string) ExtraSchemaValidator {
 	switch apiVersion {
 	case "kfd.sighup.io/v1alpha2":
--- a/internal/clusterpki/check.go
+++ b/internal/clusterpki/check.go
@@ -1,0 +1,195 @@
+// Copyright (c) 2017-present SIGHUP s.r.l All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package clusterpki
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	parserx "github.com/sighupio/furyctl/internal/parser"
+	"github.com/sighupio/furyctl/pkg/template/mapper"
+)
+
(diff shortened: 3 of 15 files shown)


internal/apis/schema.go
```diff
@@ -15,6 +15,34 @@ type ExtraSchemaValidator interface {
...
+
+// NewPKIValidatorFactory returns the PKI validator of the kind, or nil for a kind that reads no local
+// PKI. This check is separate from ExtraSchemaValidator because only `furyctl apply` and
+// `furyctl validate config` run it. The other commands must not run it. For example,
+// `furyctl create config` writes a configuration whose PKI folder does not exist yet.
+func NewPKIValidatorFactory(apiVersion, kind string) PKIValidator {
+	switch apiVersion {
+	case "kfd.sighup.io/v1alpha2":
+		switch kind {
+		case "OnPremises":
```
marcopaggioro: Myabe we can use fewer hardcoded strings using `distribution.OnPremisesKind` from `internal/distribution/compatibility.go` (either for other occurencies)

ralgozino: Yes we can, let me fix that

internal/apis/kfd/v1alpha2/immutable/public/schema.go
```diff
@@ -95,6 +95,9 @@ type SpecKubernetes struct {
 	ControlPlane SpecKubernetesControlPlane `yaml:"controlPlane"`
 	Etcd         *SpecKubernetesEtcd        `yaml:"etcd,omitempty"`
 	NodeGroups   []SpecKubernetesNodeGroup  `yaml:"nodeGroups,omitempty"`
+	// PkiPath is the folder that holds the CA certificates and keys for the control plane and etcd. The
+	// playbooks read them from the `master` and `etcd` subfolders.
+	PkiPath *string `yaml:"pkiPath,omitempty"`
```
marcopaggioro, line 100: Can you tell me why here is named `PkiPath` but for _OnPremises_ is `PkiFolder`? (so `Path` != `Folder`)

ralgozino: the variable follows the schema, for Immutable, the entry in the schema is called pkiPath and not pkiFolder. Path is more representative of the value that the user must set, i.e. a path to where the PKI files are stored.

internal/apis/schema.go
ralgozino: Fixed in [9e371a9](https://github.com/sighupio/furyctl/pull/741/commits/9e371a913308aaacc7e763b03dabc29a0172f519).

I'll open a separate PR to switch to the constants everywhere. I'll keep this PR scoped to the pki folder check.

internal/clusterpki/check.go
```diff
@@ -0,0 +1,195 @@
...
+
+				continue
+			}
+
+			return nil, fmt.Errorf("error while reading the PKI file %s: %w", path, err)
+		}
+
+		if !info.Mode().IsRegular() || info.Size() == 0 {
+			missing = append(missing, filepath.Join(dir, file))
+		}
```
marcopaggioro:
```suggestion
	var missing []string

	for _, file := range files {
		relPath := filepath.Join(dir, file)
		path := filepath.Join(absPath, relPath)

		info, err := os.Stat(path)
		if err != nil {
			if errors.Is(err, os.ErrNotExist) {
				missing = append(missing, relPath)

				continue
			}

			return nil, fmt.Errorf("error while reading the PKI file %s: %w", path, err)
		}

		if !info.Mode().IsRegular() || info.Size() == 0 {
			missing = append(missing, relPath)
		}
```

ralgozino: Am I getting it right that your suggestion is to show the user the relative path instead of the full path of the missing files in the error message?

I think printing the full path in the error is better than printing the relative path, so the user can know exactly where furyctl is trying the find the file.

Realtives path are tricky, for example furyctl supports changing the work directory with --workdir so a user running the command with a `--workdir` set could get a relative path to the workdir instead of where his shell is.

marcopaggioro: Sorry I missed the explanation. Changes should be only syntactic and not logic.
Things changed:
- Changed `missing := []string{} `to the idimatic `var missing []string`
- Extended the `path := filepath.Join(absPath, dir, file)` to
      ```relPath := filepath.Join(dir, file)
		path := filepath.Join(absPath, relPath)
		```
		to reuse relPath when needed instead of recreating twice with `filepath.Join(dir, file)`

ralgozino: ah, thanks! make sense

marcopaggioro (approved):

smerlos (approved):
With love... 💃

PR #743, opened 2026-08-07
fix(preflight): reject changes to skipped phases with --start-from

### LINKED ISSUE #657: using `--start-from` does not check for changes in other phases
### Describe the bug

Using the `--start-from` flag skips the check that changes to phases not being included have been made.

i.e.: If I made a change to the kubernetes phase and run `furyctl apply --start-from distribution` it will apply the distribution and plugin phases and the change to the kubernetes phase will be "lost".

### To Reproduce

1. Create a cluster
2. Make a change to the configuration of a phase (i.e. kubernetes)
3. Run apply with --start-from set to a phase afterwards the one you made the change (i.e. distribution)
4. See that the apply does not detect that changes to the kubernetes phase have been made and moves on, writing the update configuration to the secret in the cluster, so next time the diff won't show that the change has not been applied.

### Expected behavior

furyctl apply should stop as it does when you run an apply without the --start-from flag and changes to phases not being applied.

### Software versions

Bug is present in all latest versions of furyctl.

### Additional context

Probably the bug is in this check done in the preflight:
https://github.com/sighupio/furyctl/blob/a8c8e51b762be3fc4507c6260e42b145b54db08d/internal/apis/kfd/v1alpha2/onpremises/create/preflight.go#L204

(Linking only onprem but the same applies to ekscluster and kfddistribution)

ralgozino:
### Summary 💡

`furyctl apply --start-from <phase>` did not look for changes to the phases before the selected one. The apply skipped those changes and then wrote the new configuration to the cluster secret. As a result the change was lost, and the next `furyctl diff` did not show it as pending.

Closes #657

Relates: 
- #744

### Description 📝

The preflight check that stops an apply with changes to phases that will not be applied ran only for the `--phase` flag:

```go
if p.phase != cluster.OperationPhaseAll && !p.upgradeEnabled {
```

The value of `--start-from` never got to `PreFlight`. It arrives as an argument of `ClusterCreator.Create(startFrom, ...)`, and the constructor of the preflight got called without it. All four kinds have the same block, **so all four had the fault**.

With this PR when `--start-from` is set, furyctl rejects changes to the phases before the selected one with the new error `ErrChangesToSkippedPhases`.

The guard is now `(phase != all || startFrom != all) && !upgradeEnabled`.

### Breaking Changes 💔

None for the API or the configuration format.

Note a change of behavior: an apply with `--start-from` that has changes to an earlier phase now stops with an error. Before this fix the same command completed and lost the change.

Users who did not know about the fault can now see a new error. The correct action is to run the apply without `--start-from`, or to revert the change to the earlier phase.

### Tests performed 🧪

Manual, with a `furyctl` build from this branch, on the Immutable kind:

- [x] Created an Immutable cluster
- [x] Ran an apply after the creation with changes to the phase selected with `--phase`
- [x] Ran an apply with changes to a phase other than the one selected with `--phase`
- [x] Ran an apply with changes to a phase before the one selected with `--start-from`

All four cases behaved as expected.

### Future work 🔧

The block that calls `AssertPhaseDiffs` is identical in the four `create/preflight.go` files, and differs only by `supported.Phases()`. A shared helper can remove this duplication. This PR keeps the duplication to hold the scope to the fault.

`ValidateOperationPhase` in `cmd/apply.go` validates `--start-from` against the global list of phases, not against the phases that the kind supports. A per-kind validation can stop an invalid value earlier, with a clearer message. See the linked issue.

> [!IMPORTANT]
>

- [x] CI is green

### DIFF
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -8,8 +8,10 @@ Welcome to the latest release of `furyctl` maintained by SIGHUP by ReeVo team.
 
 ## Bug fixes 🐞
 
-TBD
+- [[#743](https://github.com/sighupio/furyctl/pull/743)] All kinds: `apply --start-from` now stops when the configuration has changes to a phase before the selected one. Before this fix furyctl skipped those changes and wrote the new configuration to the cluster, so the change was lost and the next `diff` did not show it as pending.
+
 
 ## Breaking Changes 💔
 
 None
+
--- a/internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
+++ b/internal/apis/kfd/v1alpha2/ekscluster/create/preflight.go
@@ -54,10 +54,11 @@ type PreFlight struct {
 	paths          cluster.CreatorPaths
 	force          []string
 	phase          string
+	startFrom      string
 	upgradeEnabled bool
 }
 
-func NewPreFlight(
+func NewPreFlight( //nolint:revive // ignore maximum number of arguments
 	furyctlConf private.EksclusterKfdV1Alpha2,
 	kfdManifest config.KFD,
 	paths cluster.CreatorPaths,
@@ -67,6 +68,7 @@ func NewPreFlight(
 	force []string,
 	infraOutputsPath,
 	phase string,
+	startFrom string,
 	upgradeEnabled bool,
 ) (*PreFlight, error) {
 	p := cluster.NewOperationPhase(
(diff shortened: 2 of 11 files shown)


marcopaggioro (dismissed):

stefanoghinelli (dismissed):
LGTM

ralgozino:
rebased with main to solve conflict on the unreleased.md file. Merging.

PR #745, opened 2026-08-07
feat: add `renew kubeconfigs` command

ralgozino:
### Summary 💡

Add `furyctl renew kubeconfigs` command. It renews the kubeconfig file of the admin and the kubeconfig files of the users in `spec.kubernetes.advanced.users.names` using new Ansible playbooks from the Distribution.

Closes: #575

Relates: sighupio/distribution#588

### Description 📝

```console
furyctl renew kubeconfigs                 # all of them
furyctl renew kubeconfigs admin alice     # only these two
```

The command writes the renewed files to the current directory, with the names that `furyctl apply` uses: `kubeconfig` for the admin and `<user>.kubeconfig` for each user.

**Refactor in the same PR.** `CertificatesRenewer` became `Renewer`, which has `RenewCertificates()` and `RenewKubeconfigs(users)`. One registry and one implementation for each kind serve both operations, and a private `render` method holds the template rendering that the two operations share. This removes a second parallel factory registry and about 90 duplicated lines for each kind. The behavior of `renew certificates` does not change. `cmd/renew/renew.go` now holds the bootstrap that the two subcommands share.

**Cluster kinds.** OnPremises and Immutable renew the kubeconfig files. EKSCluster and KFDDistribution give `renew is not supported for the <kind> kind`, from one shared `UnsupportedRenewer`.

Some notes for the reviewers:
- Issue #575 asked for a comma separated list. The command takes a space separated list, which is the usual form for a CLI, and accepts commas too.
- Issue #575 asked for a `secrets/users` subdirectory. The files go to the working directory instead, where `furyctl apply` already writes them, thus a renew overwrites the files that the administrator already has.
- `RunPlaybook` looks for the playbook before it calls ansible. An older SD then gives the message:
```
the distribution does not have this playbook: SD v1.35.1 does not have "renew-kubeconfigs.yaml". Upgrade the distribution to a version that has it
```

### Breaking Changes 💔

None. `renew certificates` keeps its behavior, its flags, and its output.

### Tests performed 🧪

- [x] Immutable cluster: no argument renews all the kubeconfig files
- [x] Immutable cluster: one name renews only that kubeconfig file
- [x] Immutable cluster: two names renew only those kubeconfig files, in the space form and in the comma form
- [x] Immutable cluster: an unknown name gives an error that lists the defined users
- [x] Immutable cluster: a user that I added to the configuration file gets a kubeconfig file, and the kubernetes phase is not applied
[...]

PR #749, opened 2026-08-20
fix: missing flags configuration

### LINKED ISSUE #746: flags: config values override CLI flags, and five command sections are silently ignored
### Describe the bug

The `flags:` section of `furyctl.yaml` has two defects. Both come from the order of `viper.BindPFlags` and `flags.LoadAndMergeCommandFlags` in each command's `PreRun`.

**1. A config value overrides an explicit command line flag.**

`Merger.mergeCommandFlags` guards each write with `if !viper.IsSet(viperKey)`, and the comment says: `This preserves the priority: env vars and command line flags take precedence.` The guard works only when the flags are bound first. `create` and `diff` call `LoadAndMergeCommandFlags` before `BindPFlags`, thus viper does not know the flags yet, the guard passes, and `viper.Set` wins over the flag that the user gave on the command line.

**2. Five command sections are parsed, then dropped without a message.**

`FlagsConfig` declares `validate`, `download`, `connect`, `renew` and `dump`, and `SupportedFlags` has a map for each one. `Merger.MergeIntoViper` has no `case` for them, thus they reach `default: // Unknown command, skip command-specific flags` and are discarded. The user gets no error and no warning.

| | Sections |
|---|---|
| Declared in `FlagsConfig`, parsed and validated | global, apply, delete, create, get, diff, tools, **validate, download, connect, renew, dump** |
| Handled in `MergeIntoViper` | global, apply, delete, create, get, diff, tools |

Order of the two calls in `PreRun`, for each command:

| Order | Commands | Result |
|---|---|---|
| `BindPFlags` then merge | apply, delete, get | Correct. The command line flag wins. |
| Merge then `BindPFlags` | create, diff | **Defect 1.** The config value wins. |
| Merge then `BindPFlags` | renew, validate, download, connect, dump | **Defect 2.** The section is dropped, thus defect 1 is hidden. |

### To Reproduce

Defect 1, with any `furyctl.yaml`:

1. Add this section to the file:

   ```yaml
   flags:
     diff:
       distroLocation: /tmp/ZZZ-FROM-CONFIG
   ```

2. Run `furyctl diff --distro-location /tmp/AAA-FROM-CLI`
3. The error names `/tmp/ZZZ-FROM-CONFIG`. The command ignores the path from the command line.
4. Repeat the two steps with `apply` in place of `diff`. The error names `/tmp/AAA-FROM-CLI`, which is correct.

Defect 2:

1. Add a `flags.renew.distroLocation` section to the file.
2. Run `furyctl renew certificates`.
3. furyctl downloads the distribution of the version in `furyctl.yaml`. It gives no message about the section.

### Expected behavior

1. A flag on the command line always wins over the `flags:` section of the configuration file. An environment variable also wins over the configuration file.
2. A command section that furyctl parses is also applied. If furyctl cannot apply a section, it writes a warning.

### Logs / Screenshots

```console
$ furyctl diff --distro-location /tmp/AAA-FROM-CLI
... error ... /tmp/ZZZ-FROM-CONFIG ...

$ furyctl apply --distro-location /tmp/AAA-FROM-CLI
... error ... /tmp/AAA-FROM-CLI ...
```

### Software versions

- `furyctl version` command output:

```bash
$ furyctl version
<paste your output>
```

Reproduced on `main` at `e1e34368`.

- SIGHUP Distribution version: any
- Operating system: macos 15

### Additional context

This is the cause of the confusion sibling report [#666](https://github.com/sighupio/furyctl/issues/666). That issue is not a defect: `flags.apply.distroLocation` is correct to have no effect on `renew`. The reporter could not find that out, because a move of the value to `flags.renew.distroLocation` gives no error and also has no effect.

> [!WARNING]
> Do not add the missing `case` arms alone. A `case CommandRenew:` in `MergeIntoViper`, with the current order in `PreRun`, changes defect 2 into defect 1 for `renew`. The result is the exact behavior that #666 describes. Correct the order in the same change.

For `renew`, the order is now in one place, `preRun` in `cmd/renew/renew.go` (added by #745). The other four commands each have their own `PreRun`.

Related files:

- `internal/flags/merger.go` — `MergeIntoViper` (the `switch`), `mergeCommandFlags` (the `!viper.IsSet` guard)
- `internal/flags/types.go` — `FlagsConfig`, `SupportedFlags`
- `cmd/renew/renew.go`, `cmd/validate/config.go`, `cmd/download/dependencies.go`, `cmd/connect/openvpn.go`, `cmd/dump/template.go`, `cmd/create/config.go`, `cmd/diff.go`

ralgozino:
### Summary 💡

Specifying command flags in the furyctl.yaml files was not working for some commands and being silently ignored for others. Add the logic for the missing commands.

Remove `tools` from the supported commands, it was never merged.

Closes #746

Relates:
- #619 

### Description 📝

The documentation lists validate, download, connect, renew and dump as sections of the flags field of furyctl.yaml. The merger and the validator did not know these sections. furyctl thus dropped them without a message. The validation of the configuration file also stopped with "unsupported flags command".

Add the five sections to the merger, to the validator and to the validation of the configuration file. Complete the supported flags of validate, connect, renew and dump.

Remove the tools section. That command is not in main: it lives on the branch feat/aliases-command. Its map of supported flags was empty, thus any key under flags.tools stopped every command that reads the configuration file. furyctl now rejects the section by name.

Add a test that keeps a map of supported flags for each section, and a test that merges a flag of each command section.

### Breaking Changes 💔

None

### Tests performed 🧪

- [x] `flags.renew.distroLocation` with `furyctl renew certificates`: the value is now used. Before this change furyctl dropped it without a message
- [x] `flags.renew` with `furyctl validate config`: the command now reaches the validation of the schema. Before this change it stopped with `unsupported flags command: renew`
- [x] `flags.renew.bogusFlag`: the command stops with `flag 'bogusFlag' is not supported for 'renew' command`, as the documentation says
- [x] `flags.tools.distroLocation`: the command stops with `unsupported flags command: tools`. Before this change the message named the flag instead of the section
- [x] `furyctl renew certificates --distro-location /tmp/FROM-CLI` with `flags.renew.distroLocation: /tmp/FROM-CONFIG`: the command line wins, thus #745 and the first part of #746 still hold
- [x] Regression check: the `flags.global` section still reaches `renew`, which is not a wired command section
- [x] Regression check: `flags.apply.distroLocation` with `furyctl apply` still works, thus the sections that already worked are not affected

### Future work 🔧

Seven places in the flags system each enumerate every section of the `flags` field:

| Location | Form |
|---|---|
| `Merger.MergeIntoViper` | switch, 10 arms |
| `Merger.GetSupportedFlagsForCommand` | switch, 11 arms |
| `Merger.mergeCommandFlags` | switch, 11 arms |
| `Validator.validateCommandFlags` | switch, 11 arms |
| `Validator.Validate` | if-chain, 11 blocks |
| `config.validateFlagsSection` | switch, 11 arms |
| `flags.sections_test.sectionConfig` | switch, 5 arms |

A new section needs a change in each one. #746 happened because five sections reached `FlagsConfig` and `SupportedFlags` but not the six switches. The test `TestEverySectionHasSupportedFlags` guards the result, not the cause.

Would be better if one place holds the sections. `SupportedFlags` becomes `map[string]map[string]FlagInfo`, keyed by the section name, and `FlagsConfig` gets a `Sections() map[string]map[string]any` method. The lookups then replace the switches with a map access, and the if-chain becomes a loop.

This removes about 120 lines. It also makes the fault of #746 impossible: a section that has no supported flags cannot reach the merger.

> [!IMPORTANT]
>

- [X] My branch is up-to-date with the target branch and there are no conflicts
- [X] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change
- [x] CI is green

### DIFF
--- a/docs/advanced/flags-configuration.md
+++ b/docs/advanced/flags-configuration.md
@@ -382,7 +382,7 @@ The flags configuration includes built-in validation that will **stop execution*
 - `etcd` (bool) - Create PKI only for etcd
 - `controlplane` (bool) - Create PKI only for Kubernetes control plane
 
-### Get, Diff, and Tools Command Flags
+### Get and Diff Command Flags
 
 **Get Command:**
 - `binPath` (string) - Binary path
@@ -400,11 +400,28 @@ The flags configuration includes built-in validation that will **stop execution*
 **Validate Command:**
 - `distroLocation` (string) - Distribution location
 - `distroPatches` (string) - Distribution patches location
+- `binPath` (string) - Binary path
 
 **Download Command:**
-- `binPath` (string) - Binary path  
+- `binPath` (string) - Binary path
 - `distroLocation` (string) - Distribution location
 - `distroPatches` (string) - Distribution patches location
+- `bundleOutput` (string) - Bundle tarball output path
 
-**Other Commands:**
-- `Connect`, `Renew`, `Dump` commands currently have no configurable flags - use `furyctl [command] --help` for details
+**Connect Command:**
+- `profile` (string) - OpenVPN profile name
+
+**Renew Command:**
+- `binPath` (string) - Binary path
+- `distroLocation` (string) - Distribution location
+- `skipDepsDownload` (bool) - Skip dependencies download
+- `skipDepsValidation` (bool) - Skip dependencies validation
+- `airgapBundle` (string) - Air-gapped bundle path
+- `forceExtract` (bool) - Force bundle re-extraction
+
+**Dump Command:**
+- `distroLocation` (string) - Distribution location
+- `distroPatches` (string) - Distribution patches location
+- `dryRun` (bool) - Dry run
+- `noOverwrite` (bool) - Do not overwrite existing files
+- `skipValidation` (bool) - Skip validation
--- a/docs/releases/unreleased.md
+++ b/docs/releases/unreleased.md
@@ -11,6 +11,7 @@ Welcome to the latest release of `furyctl` maintained by SIGHUP by ReeVo team.
 
 - [[#743](https://github.com/sighupio/furyctl/pull/743)] All kinds: `apply --start-from` now stops when the configuration has changes to a phase before the selected one. Before this fix furyctl skipped those changes and wrote the new configuration to the cluster, so the change was lost and the next `diff` did not show it as pending.
 - [[#746](https://github.com/sighupio/furyctl/issues/746)] All kinds: a flag that you give on the command line now has precedence over the same flag in the `flags` section of `furyctl.yaml`. The documented priority is `furyctl.yaml` < environment variable < command line. Before this release the commands `diff`, `create config`, `get cluster-info` and `get upgrade-paths` read the configuration file first, thus the configuration file had the precedence. If you want the value of the configuration file, do not give the flag on the command line.
+- [[#746](https://github.com/sighupio/furyctl/issues/746)] All kinds: the `validate`, `download`, `connect`, `renew` and `dump` sections of the `flags` field of `furyctl.yaml` now work. The documentation lists these sections. Before this release furyctl dropped them. The validation of the configuration file also stopped with `unsupported flags command`. furyctl does not accept a `tools` section any more, because it has no `tools` command.
 
 
 ## Breaking Changes 💔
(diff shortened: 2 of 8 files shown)


stefanoghinelli (approved):
Looking good

PR #752, opened 2026-08-21
fix(flags): align the supported flags table with the commands

### LINKED ISSUE #750: flags: the supported-flags table does not match the flags the commands register
### Describe the bug

The `SupportedFlags` table in `internal/flags/types.go` does not agree with the flags that the commands register. There are two faults, and each one has the opposite effect for the user.

**Phantom entries.** The table lists a flag that no command has. furyctl accepts it in the `flags` section of `furyctl.yaml` and validates it, then nothing reads the value. The user gets no message.

**Missing entries.** A command has a flag, but the table does not list it. furyctl stops with `flag 'X' is not supported for 'Y' command`, although the flag exists.

| Section | Phantom: in the table, no such flag | Missing: real flag, absent from the table |
|---|---|---|
| `create` | `provider` | `--api-version`, `--kind`, `--distro-location`, `--distro-patches` |
| `delete` | `autoApprove`, `startFrom` | `--force`, `--skip-deps-download`, `--skip-deps-validation`, `--vpn-auto-connect` |
| `get` | — | `--format`, `--from`, `--kind` |
| `diff` | — | `--skip-deps-download` |

The `create` section has it backwards in both directions at the same time: the real `--kind` flag is rejected, and the `provider` flag, which does not exist, is accepted.

### To Reproduce

1. Put this section in any `furyctl.yaml`:

   ```yaml
   flags:
     create:
       kind: OnPremises
   ```

2. Run `furyctl validate config`.
3. The command stops with `flag 'kind' is not supported for 'create' command`. The flag `furyctl create config --kind` exists.
4. Replace `kind: OnPremises` with `provider: aws`, then run the command again. furyctl accepts the section. No command reads `provider`.

### Expected behavior

Each section of the table lists the flags of its commands, and only those flags. A test compares the table with the command tree, thus the two stay together when a command gains a flag.

### Logs / Screenshots

```console
$ furyctl validate config     # with flags.create.kind
... fatal validation error for create.kind: flag 'kind' is not supported for 'create' command ...

$ furyctl validate config     # with flags.create.provider
... no message about the flags section ...
```

### Software versions

present since the introduction of flags in the configuration file.

### Additional context

The `global` section does not list `--https`. That flag is deprecated in favor of `--git-protocol`, thus its absence looks deliberate. This report does not ask for it.

A probe found these gaps. It walks the command tree of cobra from `cmd.NewRootCmd()` and compares each section of `GetSupportedFlags()` with the flags of the commands of that section. It reports the phantom and the missing names for every section. A test of this shape is the fix for the second part of the expected behavior: it fails when a command gains a flag that the table does not list.

The five sections that #749 added — `validate`, `download`, `connect`, `renew` and `dump` — have no gap. The gaps are in the four sections that were present before #749.

Related: #746, #749.

ralgozino:
### Summary 💡

The `flags` section of `furyctl.yaml` validates each flag against a table of supported flags. The table did not match the commands, in both directions: 8 flags that the commands have were absent, and 5 flags that no command has were present. This PR makes the table agree with the commands and adds a test that keeps them in agreement.

Closes #750

Relates:
- #746
- #749
- #751

### Description 📝

**The 8 flags that furyctl now accepts.** 

- `delete` gains `force`, `skipDepsDownload`, `skipDepsValidation` and `vpnAutoConnect`.
- `get` gains `format`, `from` and `kind`.
- `diff` gains `skipDepsDownload`.

Before this PR furyctl stopped with `flag 'X' is not supported`, although the command has the flag.

Note that `delete cluster --force` is a boolean, but `apply --force` takes a list. The table gives each one its own type, and the validation rejects the wrong one.

**The 5 flags that furyctl does not accept any more.**

These are `create.provider`, `create.name`, `create.version`, `delete.startFrom` and `delete.autoApprove`. No command has them. furyctl accepted them and then did nothing with them.

`create config` creates the configuration file, thus it cannot read flags from one: it stops when the file is already present. Only `create pki` reads the `create` section. For this reason `cmd/create/config.go` does not load the section any more. The global flags still apply, because `cmd/root.go` merges them.

Added a test `internal/flags/table_test.go` that compares the table with the flags that the commands declare. A command that gains a flag makes the test fail until the table has the flag, and a table entry that no command has also makes it fail. The test names the flags that it excludes on purpose, and gives the reason for each one.

The reference lists in `docs/advanced/flags-configuration.md` were out of date in 5 sections, and two examples used flags that no command has.

### Breaking Changes 💔

furyctl does not accept `create.provider`, `create.name`, `create.version`, `delete.startFrom` and `delete.autoApprove` in the `flags` section any more. Remove them from your `furyctl.yaml`. These flags were not working anyway.

Until you remove them, **every** furyctl command that you run in that directory stops, also the commands that do not read those sections:

```
$ furyctl get supported-versions --kind OnPremises
ERRO fatal validation error for create.name: flag 'name' is not supported for 'create' command.
```

The reason is that furyctl validates the complete `flags` section at each start. The message names the section and the flag to remove.

### Tests performed 🧪

- [x] Verified the 8 flags that were absent: each one is accepted in the `flags` section of a `furyctl.yaml`, for `delete`, `get` and `diff`
- [x] Verified the 5 flags that were removed: each one is rejected, and the message names the flag and the section
- [x] Verified that the flags which must keep working are still accepted: `create.path`, `create.etcd`, `apply.startFrom` and `delete.phase`
- [x] Verified the two types of `force`: `delete.force: true` is accepted and `delete.force: ["upgrades"]` is rejected, `apply.force: ["upgrades"]` is accepted and `apply.force: true` is rejected
- [x] Verified that `apply.force` still rejects an invalid option, and that the message lists the valid ones
- [x] Verified the blast radius: a removed flag in `./furyctl.yaml` stops an unrelated command (`get supported-versions`), and the same file on `main` does not
- [x] Verified `furyctl create config`: it writes the file, it still stops when the file is already present, and `flags.global.debug: true` in a neighbouring `furyctl.yaml` still turns on the debug output
- [x] Compared the reference lists in the documentation with the table: they agree
- [x] Parsed every YAML example in `flags-configuration.md` and checked each flag against the table: all valid, except the counter-example that the documentation marks as a fatal error

### Future work 🔧

The `flags` section is validated in full at each start, thus a stale flag for one command
stops all the other commands. A warning for a section that the command does not read could
be friendlier than a fatal error.

> [!IMPORTANT]
>

- [x] CI is green


internal/flags/table_test.go
```diff
@@ -0,0 +1,140 @@
...
+	// `https` is deprecated in favor of `git-protocol`.
+	"https",
+}
+
+var flagName = regexp.MustCompile(`--([a-z][a-z0-9-]*)`)
+
+// TestTableMatchesTheCommands keeps the supported flags and the commands together. It fails when a
+// command gains a flag that the table does not list, and when the table lists a flag that no
+// command has.
+func TestTableMatchesTheCommands(t *testing.T) {
```
stefanoghinelli, line 66: It checks that flag names in GetSupportedFlags() match the flags declared by the commands, but it does not compare their types. A mismatch such as `delete.force` being a bool and `apply.force` being a string slice would still pass this test, even though it would cause an error at runtime. WDYT?

stefanoghinelli (commented):
1 question, otherwise LGTM

internal/flags/table_test.go
ralgozino: makes sense, let me try that

ralgozino: type check implemented, nice catch!

stefanoghinelli (approved):

PR #754, opened 2026-08-28
fix(on-prem): resume skipped worker nodes during upgrade with `--skip-nodes-upgrade` flag

### LINKED ISSUE #718: on-prem: `--upgrade` flag gives no warning when furyctl silently runs a reconciliation instead of a real upgrade
### Describe the bug

furyctl decides whether `--upgrade` means a real kubeadm upgrade or a reconciliation (`create-playbook.yaml`) by diffing spec.distributionVersion in `furyctl.yaml` against the state persisted in the furyctl-config Secret in the cluster. If it finds no diff there, apply `--upgrade` runs `create-playbook.yaml` instead of the real upgrade playbooks (`55.upgrade-control-plane.yml`, `56.upgrade-worker-nodes.yml`), with no warning and no confirmation prompt, even though `--upgrade` was passed explicitly. 

In 

https://github.com/sighupio/furyctl/blob/58b72e3796df60d91d61d5a34398fc2007a8fd85/internal/apis/kfd/v1alpha2/onpremises/creator.go#L131-L134

`c.upgrade` is the flag passed by the user, `upgr.Enabled` is what furyctl decides to do (set inside `PreUpgrade.Exec()` only when the diff is present), and nothing ever compares the two.

The deeper issue is not that furyctl fails to warn, it's that it never persists that a `--skip-nodes-upgrade` run left the worker upgrade incomplete. `creator.go` (around L337) deletes the `furyctl-upgrade-state` ConfigMap whenever `upgr.Enabled` is true, regardless of whether nodes were actually skipped, and `PreUpgrade.Exec()` only sets `upgr.From`/`upgr.To` when a live diff exists. So after a staged run, nothing in the cluster records that the workers still need `56.upgrade-worker-nodes.yml`, and the next run has no way to know.

### To Reproduce

1. On an existing cluster, bump distributionVersion in `furyctl.yaml` and run `furyctl apply --upgrade --skip-nodes-upgrade`. This upgrades the control plane only and persists the new distributionVersion to the furyctl-config Secret.
2. Run `furyctl apply --upgrade` again, intending to bring the workers up to the same version.
3. With no warning prompt, runs `create-playbook.yaml` on the workers instead of `56.upgrade-worker-nodes.yml` and kubeadm   is not run.

stefanoghinelli:
### Summary 💡

Fix `furyctl apply --upgrade` on OnPremises when worker nodes were skipped. It now upgrades the worker nodes that have not been upgraded yet, one at a time, with the same procedure used by `--upgrade-node`.

Closes: https://github.com/sighupio/furyctl/issues/718

### Description 📝

OnPremises supports upgrading the control plane first and the worker nodes at a later time.

```bash
furyctl apply --upgrade --skip-nodes-upgrade
furyctl apply --upgrade-node <node name>
```

Before this change, `furyctl apply --upgrade` after the control plane upgrade selected `create-playbook.yaml` for the worker nodes because the target configuration was already saved. furyctl now saves the source and target Distribution versions and the result for each worker node. It can continue the upgrade after the configuration is saved. `furyctl apply --upgrade` asks for confirmation and upgrades only the remaining or failed worker nodes one at a time. Dry run reports which nodes would be upgraded without asking for confirmation.

furyctl saves the result after every worker node in the same `furyctl-upgrade-state` ConfigMap. A later `apply --upgrade` retries only failed and remaining worker nodes. While workers remain, furyctl stops an ordinary apply and partial upgrade options. `--force upgrades` and `--force all` allow partial upgrade options to continue with a warning. furyctl stops when saved progress cannot be read and reads it only after preflight verifies the target cluster. Existing upgrade state remains compatible.

### Breaking Changes 💔

No schema changes, but some commands now stop instead of performing a potentially unsafe apply. A plain `furyctl apply` stops while worker nodes remain. furyctl also rejects `--phase`, `--start-from` and `--post-apply-phases` unless `--force upgrades` or `--force all` is set, and rejects `--upgrade-node <node> --skip-nodes-upgrade`. Repeating `furyctl apply --upgrade --skip-nodes-upgrade` after the control plane upgrade completes successfully without running playbooks.

### Tests performed 🧪

- [x] OnPremises cluster upgrade from `v1.34.1` to `v1.35.1` on a cluster with three control plane nodes and three worker nodes. The test covered the control plane upgrade with workers skipped, blocked commands, dry run without interactive input, a direct worker upgrade, a repeated worker upgrade, confirmation and the upgrade of the remaining worker nodes.
- [x] Verified `--phase`, `--start-from` and `--post-apply-phases` while worker nodes were pending. They stopped without force and continued with a warning when using `--force upgrades` or `--force all`.
- [x] A testing PodDisruptionBudget blocked the drain of one worker node. The next `apply --upgrade` retried only that worker and the worker nodes that had not been upgraded yet.
- [x] Retried an incomplete control plane upgrade while continuing to skip the worker nodes.
- [x] KFDDistribution normal upgrade smoke test.

### Future work 🔧

- Update the doc site
- Apply the same fix to Immutable

> [!IMPORTANT]
>

- [x] I've updated the documentation for the OnPremises upgrade workflow
- [x] My branch is current with the target branch and there are no conflicts
- [x] I've considered all the different cluster kinds that may be affected by this change
- [x] CI is green


internal/apis/kfd/v1alpha2/onpremises/create/kubernetes.go
```diff
@@ -243,6 +290,26 @@ func (k *Kubernetes) preKubernetes(
 
 		if k.upgrade.Enabled {
 			upgradeState.Phases.PreKubernetes.Status = upgrade.PhaseStatusSuccess
+
+			if k.skipNodesUpgrade && len(k.workerNodes) > 0 {
+				hasScript, err := k.upgrade.HasScript("pre-kubernetes")
```
ralgozino: what if the upgrade path being applyed does not include upgrades for the kubernetes phase? for example, is a patch version with only changes to the distribution phase and the user runs `furyctl apply --upgrade` without the phase flag. Will this check get triggered?

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -133,19 +147,34 @@ func (*ClusterCreator) GetPhasePath(phase string) (string, error) {
 
 func (c *ClusterCreator) Create(startFrom string, _, podRunningCheckTimeout int) error {
 	upgr := upgrade.New(c.paths, string(c.furyctlConf.Kind))
+	var workerNodes []string
+	if c.skipNodesUpgrade && c.upgrade {
+		if c.phase != cluster.OperationPhaseAll || startFrom != "" || len(c.postApplyPhases) > 0 {
+			return fmt.Errorf(
+				"%w: --skip-nodes-upgrade requires a full apply; do not combine it with "+
+					"--phase, --start-from, or --post-apply-phases",
```
ralgozino: Is this constraint new? Why block, for example, an upgrade of just the kubernetes phase for the control plane? I think you were able to do this before.

is this a breaking change?

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -323,64 +438,388 @@ func (c *ClusterCreator) Create(startFrom string, _, podRunningCheckTimeout int)
...
+	upgradeState *upgrade.State,
+	renderedConfig map[string]any,
+) error {
+	if err := c.storeTargetConfig(renderedConfig); err != nil {
+		return err
 	}
 
+	upgradeState.StagedWorkers.ReadyForResume = true
+	if err := c.upgradeStateStore.Store(upgradeState); err != nil {
+		return fmt.Errorf("error while marking staged worker upgrade ready to resume: %w", err)
```
ralgozino, line 535: what would be the recovery plan if this error occurs? we will have the updated config stored but readyforresume in false

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -191,6 +220,44 @@ func (c *ClusterCreator) Create(startFrom string, _, podRunningCheckTimeout int)
 	if err != nil {
 		return fmt.Errorf("error while executing preflight phase: %w", err)
 	}
+	var existingUpgradeState *upgrade.State
+	if status.ClusterExists && !c.dryRun {
```
ralgozino: this dryRun guard will skip all the logic for detecting the ugprade process status, so a dryRun for an ongoing upgrade will say that a full upgrade would have taken place, wouldn't it?

ralgozino (changes requested):
In addition to the comments below:

- do the changes impact the `get cluster-info` command? should it show for example the pending nodes to upgrade
- maybe I missed it, when the user updates the control-plane only, there's no message telling the user what command to run to update the remaining nodes. When updating a single node it would be nice to show a list of the remaining nodes if we have the info already.

stefanoghinelli:
Thanks for the hints, will do.

> * do the changes impact the `get cluster-info` command? should it show for example the pending nodes to upgrade
> * maybe I missed it, when the user updates the control-plane only, there's no message telling the user what command to run to update the remaining nodes. When updating a single node it would be nice to show a list of the remaining nodes if we have the info already.

These are also reasonable changes to make.

internal/apis/kfd/v1alpha2/onpremises/create/kubernetes.go
stefanoghinelli: Fixed

internal/apis/kfd/v1alpha2/onpremises/creator.go
stefanoghinelli: Fixed

internal/apis/kfd/v1alpha2/onpremises/creator.go
stefanoghinelli: Yes, done

internal/apis/kfd/v1alpha2/onpremises/creator.go
stefanoghinelli: Yes, it was new and it would have been a breaking change. I removed the constraint, so upgrading only the Kubernetes phase while skipping worker nodes works as before.

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -589,8 +598,20 @@ func (c *ClusterCreator) stagedUpgradeDecision(
...
+			if err := validateStagedTransition(upgradeState, changes); err != nil {
+				return stagedUpgradeProceed, err
+			}
+
+			return stagedUpgradeFinalize, nil
+		}
+
 		return stagedUpgradeProceed, rejectIncompleteStagedUpgrade(upgradeState, changes)
 	}
+	phaseScoped := c.phase != "" && c.phase != cluster.OperationPhaseAll
```
ralgozino: `cluster.OperationPhaseAll` is already an empty string (`""`), so this line could be simplified to:
```suggestion
	phaseScoped := c.phase != cluster.OperationPhaseAll
```

or just use `c.phase != cluster.OperationPhaseAll` below

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -323,64 +451,413 @@ func (c *ClusterCreator) Create(startFrom string, _, podRunningCheckTimeout int)
...
+				return stagedUpgradeProceed, err
+			}
+
+			return stagedUpgradeFinalize, nil
+		}
+
+		return stagedUpgradeProceed, rejectIncompleteStagedUpgrade(upgradeState, changes)
+	}
+	phaseScoped := c.phase != "" && c.phase != cluster.OperationPhaseAll
+	if c.skipNodesUpgrade || phaseScoped || startFrom != "" || len(c.postApplyPhases) > 0 {
```
ralgozino: A phase-scoped apply like `furyctl apply --upgrade --phase distribution` exits 0 and does not run the distribution phase.

To see this:

1. Run `furyctl apply --upgrade --skip-nodes-upgrade`. The control plane upgrades. The worker nodes stay pending.
2. Run `furyctl apply --upgrade --phase distribution`.

furyctl prints one line about the pending worker nodes and exits 0. The distribution phase does not run.

The cause is this condition. It returns `stagedUpgradeNoop`, and the `case stagedUpgradeNoop` arm returns `nil` from `Create` before `executePhase`.

Command line flags are not the only route here. You can also set `phase`, `startFrom` and `postApplyPhases` in furyctl.yaml. If a user sets `postApplyPhases` in that file, this condition is always true. Every `furyctl apply --upgrade` then exits 0 with no work done. `logStagedWorkerNextSteps` names `furyctl apply --upgrade` as the next step, so the user runs the same command again.

`--skip-nodes-upgrade` belongs in this branch. A second run with that flag has no work to do.

For `phase`, `startFrom` and `postApplyPhases`, two options:

- Remove them from the condition, so the run continues to `stagedUpgradeProceed`. Pending worker nodes do not make a distribution re-apply unsafe.
- Keep the block and return the error instead of `nil`. A non-zero exit code tells a pipeline that furyctl did no work.

If you decide to keep the block note that `OperationPhaseAll` is already an empty string (`""`)

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -323,64 +451,413 @@ func (c *ClusterCreator) Create(startFrom string, _, podRunningCheckTimeout int)
...
+	}
+
+	// Do not list worker node names because a cluster can have hundreds of workers.
+	message := fmt.Sprintf(
+		"\nResuming staged upgrade %s to %s for %d pending worker nodes.",
+		upgradeState.Transition.From,
+		upgradeState.Transition.To,
+		len(nodes),
+	)
+	confirm, err := cluster.AskConfirmationWithMessage(
```
ralgozino: `furyctl apply --upgrade --dry-run` asks the user for confirmation. If stdin is not a terminal, the command fails.

To see this:

1. Run `furyctl apply --upgrade --skip-nodes-upgrade`. The control plane upgrades. The worker nodes stay pending.
2. Run `furyctl apply --upgrade --dry-run < /dev/null`.

furyctl stops with this error:

`error while creating cluster: staged worker upgrade: error asking for confirmation: error reading user input: error reading from stdin: EOF`

The gate at line 218 now lets a dry run through, which is the correct change. A dry run therefore reaches `resumeStagedWorkerBatch`, and this call has no dry-run guard. `AskConfirmationWithMessage` reads `os.Stdin`, and `Prompter.Ask` returns an error at end of file.

In a terminal furyctl waits for `yes` before it prints `Would upgrade %d worker nodes (dry-run mode)`. `--force upgrades` skips the prompt, but a dry run must not need a force flag.

Suggested change: wrap lines 790 to 799 in `if !c.dryRun`.

internal/apis/kfd/v1alpha2/onpremises/create/kubernetes.go
```diff
@@ -124,6 +126,51 @@ func (k *Kubernetes) SetUpgrade(upgradeEnabled bool) {
 	k.upgrade.Enabled = upgradeEnabled
 }
 
+// UpgradeWorkerNodes upgrades the requested worker nodes.
+func (k *Kubernetes) UpgradeWorkerNodes(
+	nodes []string,
+	onResult func(node string, status upgrade.PhaseStatus) error,
+) error {
+	if err := k.prepare(); err != nil {
```
ralgozino: `prepare()` runs before the `k.dryRun` check below it. A dry run therefore renders the templates and runs `ansible all -m ping` against every host.

If we want to skip the ansible ping we should move the `k.dryRun` block above this call. I'm not sure the ping should run or not on a dryrun. Either case is OK

ralgozino (changes requested):

internal/apis/kfd/v1alpha2/onpremises/creator.go
```diff
@@ -323,64 +451,413 @@ func (c *ClusterCreator) Create(startFrom string, _, podRunningCheckTimeout int)
...
+			return stagedUpgradeFinalize, nil
+		}
+
+		return stagedUpgradeProceed, rejectIncompleteStagedUpgrade(upgradeState, changes)
+	}
+	phaseScoped := c.phase != "" && c.phase != cluster.OperationPhaseAll
+	if c.skipNodesUpgrade || phaseScoped || startFrom != "" || len(c.postApplyPhases) > 0 {
+		return stagedUpgradeNoop, nil
+	}
+	if len(changes) != 0 {
```
ralgozino, line 617: **Every other unsafe gate in this file reads `c.force`. The staged-upgrade rejections do not.**

Lines 472, 791, 907 and 928 all call `cluster.IsForceEnabledForFeature`. This return and the other `errStagedUpgrade` returns call nothing. `--force upgrades` has no effect here, and `--force all` has no effect either. `IsForceEnabledForFeature` matches `all` inside the function, but no rejection on this path calls that function.

The user must therefore delete `cm/furyctl-upgrade-state` in `kube-system` by hand. No message names that ConfigMap.

Suggested change: `stagedUpgradeDecision` has the receiver, so it can call `cluster.IsForceEnabledForFeature(c.force, cluster.ForceFeatureUpgrades)` directly. No signature change is necessary. If force is set, write a warning and continue instead of this return.

internal/apis/kfd/v1alpha2/onpremises/creator.go
stefanoghinelli: Already considered the first, but it doesn't fit because --post-apply-phases still performs a normal apply before running the selected post apply phase. With pending workers, runs create-playbook.yaml without running   kubeadm upgrade node. The second option seems to be the safest.

internal/apis/kfd/v1alpha2/onpremises/creator.go
stefanoghinelli: Fixed as suggested

internal/apis/kfd/v1alpha2/onpremises/creator.go
stefanoghinelli: Done and rested

ralgozino (dismissed):
LGTM. Well done!

PR #756, opened 2026-09-07
feat(create): add secrets subcommand

ralgozino:
### Summary 💡

The fields of `furyctl.yaml` that hold a secret tell you which command makes their value, and you run each one by hand: `head -c32 /dev/urandom | base64` for two of the Pomerium secrets, an `openssl ecparam` invocation for its signing key. A cluster with SSO needs four values, and nothing tells you which fields are still empty.

`furyctl create secrets` writes them for you, one file for each secret, and tells you how to read those files from the configuration file.

Relates: 
- sighupio/distribution#594
- sighupio/docs#202

### Description 📝

<details>
<summary>New create secrets command's help message</summary>

```bash
furyctl create secrets --help
Creates the random secrets that the configuration file of a cluster needs. For example, Pomerium needs four secrets when the authentication provider type is "sso".

furyctl writes one file for each secret, in a folder with the name of the component. The configuration file reads the value of each file with the "{file://<path>}" notation. The command prints the fields to add to the configuration file.

With no argument, furyctl creates the secrets of the components that the configuration file uses. When there is no configuration file, furyctl creates the secrets of every component. You can name the components to create only some of them, for example "furyctl create secrets pomerium keepalived".

The components are: pomerium, gangplank, basic-auth, minio-logging, minio-monitoring, minio-tracing, keepalived, keepalived-controlplane, etcd-encryption, haproxy-stats.

On the OnPremises and Immutable kinds furyctl also asks whether the cluster encrypts the secrets of etcd at rest. It writes that configuration only after an answer of "yes", or when you name the component. It asks nothing with the --no-tty flag.

A file that is already there keeps its value: furyctl never writes a secret twice. Thus you can run the command again after you add a module to the configuration file.

Usage:
  furyctl create secrets [COMPONENT...] [flags]

Examples:
  furyctl create secrets                            Create the secrets that furyctl.yaml needs
  furyctl create secrets pomerium                   Create only the secrets of Pomerium
  furyctl create secrets --path ./mysecrets         Write the files in another folder
  furyctl create secrets --update-config            Also write the references in furyctl.yaml
  furyctl create secrets --config mycluster.yaml    Read another configuration file


Flags:
  -c, --config string   Path to the configuration file (default "furyctl.yaml")
  -h, --help            help for secrets
  -p, --path string     path where to save the created secrets. furyctl creates one subfolder for each component (default "./secrets")
      --update-config   write the reference to each file in the configuration file. furyctl edits only the lines of those fields, thus the comments and the format of the rest of the file do not change. A field that already holds another value does not change

Global Flags:
  -D, --debug                       Enables furyctl debug output. This will greatly increase the verbosity. Debug logs and additional logs are also always written to the log file. See the --log flag.
  -d, --disable-analytics           Disable analytics
  -g, --git-protocol git-protocol   Download repositories using the given protocol. Use when SSH traffic is being blocked or when SSH client has not been configured
                                    set the GITHUB_TOKEN environment variable with your token to use authentication while downloading, for example for private repositories.
                                    Options are: ssh, https (default https)
  -H, --https                       DEPRECATED: by default furyctl uses https protocol to download repositories. See --git-protocol flag (default true)
  -l, --log string                  Path to a file or folder where to write logs to. Set to 'stdout' write to standard output. Target path will be created if it does not exists. Path is relative to --workdir. Default is '<outdir>/.furyctl/furyctl.<timestamp>-<random number>.log'
  -T, --no-tty                      Disable TTY making furyctl's output more friendly to non-interactive shells by disabling animations and colors
  -o, --outdir string               Path where to create the ".furyctl" data directory. Default is the user's home. Path is relative to --workdir
  -w, --workdir string              Switch to a different working directory before executing the given subcommand. NOTE: this will affect all the paths passed, including other flags like outdir and log, for example
```

</details>

`furyctl create secrets [COMPONENT...]` writes one file for each secret in `./secrets/<component>/`, then prints the fields to add to `furyctl.yaml` with the `{file://<path>}` notation.

An option `--update-config` flag writes those fields for you, `--path` chooses another folder.

Supported `COMPONENT`s:
- `pomerium`
- `gangplank`
- `basic-auth`
- the MinIO root password of the logging, monitoring and tracing modules
- `keepalived`
- `keepalived-controlplane` (immutable)
- `haproxy-stats`
- `etcd-encryption`.

With no argument the command creates only the secrets that your configuration file needs, and it works without a configuration file too.

**A file is never written twice.** The `O_EXCL` flag makes the check and the creation one operation, and the value is generated before the file exists, so a failure cannot leave an empty file that a later run would keep. A second run reports which files it kept.

**Permissions.** New files get 0600 and new folders 0700. furyctl changes the permissions of no file that it did not write, and reports the folder and the files that other accounts of the machine can read, with the `chmod` that closes them.

**`--update-config` does not reformat your configuration file.** It parses the file to find where each field is, then edits the text of those lines. Comments, empty lines, quoting style and multi-line values are untouched. A field that already holds another value is reported, never replaced.

**The etcd encryption at rest is opt-in by a question.** The configuration file does not have a clear indicator of if etcd encryption is enabled, because the presence of the field is the switch, so no run can enable a control plane feature on its own. With an answer of `yes` furyctl writes an `EncryptionConfiguration` object with a new `secretbox` key and the `identity` provider last, and the API server picks it up with the kubernetes phase of an apply. `--no-tty`, a closed standard input, and a kind without the field all mean no question.

`furyctl create pki` also tells you now to handle the files that it writes with care, with the same message.

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] `create secrets` on an `OnPremises` configuration with `type: sso`: four Pomerium files, and the printed fields match the schema
- [x] `--update-config` on a configuration file with comments, empty lines, a trailing comment on the target field, and an `oidcTrustedCA` block scalar: only the new fields appear in the diff, and the file still parses
- [x] `--update-config` where `furyctl.yaml` is a symbolic link: the link stays a link and the file at the end of it holds the new fields
- [x] Second run of both forms: no file written again, no field written again, and the command says so
- [x] Values checked one by one: the cookie and shared secrets decode to 32 bytes, the signing key opens with `openssl ec`, the Keepalived passphrase is 8 characters, the `EncryptionConfiguration` parses and its key decodes to 32 bytes
- [x] No configuration file: each component except the encryption at rest
- [x] A named component, a name twice, and an unknown name
- [x] The encryption question with `yes`, with `no`, with `--no-tty`, with a closed standard input, with the component named, and on an `EKSCluster` configuration
- [x] `Immutable` with both Keepalived clusters: two files, two different passphrases, at the two fields of that kind
- [x] `Immutable` with MinIO for Mimir and an external endpoint for Tempo: only the MinIO password of the monitoring module
- [x] A fresh clone of a repository that holds the secrets: the command reports the folder and the four files that other accounts can read, keeps each value, and changes no permissions
- [x] The same clone with the folders closed by hand: the command reports the files only. With everything closed: no message
- [x] `--workdir` from another folder: the files and the configuration file are those of the working directory, each message names an absolute path, and the references stay relative to the configuration file
- [x] A configuration file that holds no fields, with `--update-config`: the command stops and the file stays as it is

### Future work 🔧

- `furyctl create secrets` could get a rotation mode. It needs a plan for the API server restart and for the re-encryption of the data that etcd holds, thus it is not part of this PR.
- `validate config` could report a secret field that is empty, or a `{file://}` reference whose file is absent, in the same way that it reports an absent PKI folder today.

- [x] CI is green


internal/secrets/secrets.go
```diff
@@ -0,0 +1,510 @@
...
+				{
+					Name: "SHARED_SECRET",
+					Path: "spec.distribution.modules.auth.pomerium.secrets.SHARED_SECRET",
+					New:  key,
+				},
+				{
+					Name: "SIGNING_KEY",
+					Path: "spec.distribution.modules.auth.pomerium.secrets.SIGNING_KEY",
+					New:  signingKey,
+				},
```
stefanoghinelli, line 137: What happens when the configuration targets an older supported distribution version? Like distributionVersion: v1.25.8 and sso enabled, --update-config adds SIGNING_KEY, but that version's schema does not define it and has additionalProperties: false. Should the list of secrets depend on spec.distributionVersion?

internal/secrets/secrets.go
```diff
@@ -0,0 +1,510 @@
...
+			component, found := lo.Find(all, func(c Component) bool { return c.Name == name })
+			if !found {
+				return nil, fmt.Errorf("%w %q: furyctl creates the secrets of %s",
+					ErrUnknownComponent, name, strings.Join(Names(), ", "))
+			}
+
+			selected = append(selected, component)
+		}
+
+		return resolve(selected, cfg), nil
```
stefanoghinelli, line 287: Should explicitly named components still be checked against the configuration kind? furyctl create secrets etcd-encryption --update-config on an EKSCluster currently adds spec.kubernetes.advanced.encryption, and the EKS schema then rejects the configuration with additionalProperties 'advanced' not allowed. The automatic encryption path checks the kind, but this path seems to bypass that check.

internal/secrets/secrets.go
```diff
@@ -0,0 +1,510 @@
...
+	value, err := generate()
+	if err != nil {
+		return false, err
+	}
+
+	f, err := os.OpenFile(file, os.O_WRONLY|os.O_CREATE|os.O_EXCL, filePermissions)
+	if err != nil {
+		if errors.Is(err, os.ErrExist) {
+			return false, nil
+		}
```
stefanoghinelli, line 578: Should we verify that an existing target is a readable, non-empty regular file before treating it as a secret? I tried this with an empty file and with a directory named password: both were reported as “kept”, and --update-config wrote the reference. The empty file remained empty, while validation later failed.

internal/secrets/config.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil, fmt.Errorf("error while reading the configuration file %s: %w", path, err)
+	}
+
+	root := map[string]any{}
+
+	if err := yaml.Unmarshal(content, &root); err != nil {
+		return nil, fmt.Errorf("error while parsing the configuration file %s: %w", path, err)
+	}
+
+	return &Config{root: root}, nil
```
stefanoghinelli: This should maybe expand dynamic values in the same way as the regular configuration loader. With provider.type: "{env://AUTH_TYPE}" and AUTH_TYPE=sso, the command reports that no secrets are needed. Replacing the reference with the static value sso selects the four Pomerium secrets.

stefanoghinelli (changes requested):

internal/secrets/secrets.go
ralgozino: It's a valid point. The real answer to that question is the schema version should be bumped when it has relevant changes, and then we could use that to version this kind of checks.

But the truth is that we are not currently bumping the schema version for reasons™.

Until we do that, I think the most sane thing to do is to have some version check here - like min and max distribution version for the field.

For the particular exmaple on Pomerium's key: I don't think there's a real risk of someone on that old version using this new `create secrets` command against that cluster configuration, but the point stands.

Thoughts?

internal/secrets/secrets.go
ralgozino: Yes they should, thanks for pointing it out

internal/secrets/secrets.go
ralgozino: Yeap, that's a bug, nice catch

internal/secrets/config.go
ralgozino: Nice catch, will fix

internal/secrets/secrets.go
stefanoghinelli: +1 to using the schema version

ralgozino: I took a different route, furyctl now resolves the real public schema for the configuration's kind and version, and asks it whether each field exists — by putting a value at the field and validating it. So `create secrets` and `validate config` can never disagree. There's no min and max versions table to maintain.

A new secret in a future SD version still needs a hand-written entry in `All(`) — the field path and its generator — because nothing in the schema marks a field as one furyctl should generate a value for.

internal/secrets/secrets.go
ralgozino: Fixed. Named components now go through the same check. `furyctl create secrets etcd-encryption --update-config` on an EKSCluster exits 1 with the cluster has no field for the component: etcd-encryption, writes no file and leaves furyctl.yaml untouched — verified.

internal/secrets/secrets.go
ralgozino: Fixed. A path that is not a regular file, or a file with no value in it, is now an error instead of "kept", so no reference gets written for it.

Example:
```
$ furyctl create secrets --config immutable.yaml --update-config
INFO Reading the schema of the cluster...
ERRO cannot read the value of a field of the configuration file: spec.infrastructure.loadBalancers.keepalived.passphrase. Give each field a value that furyctl can read, or take it out of /tmp/createsecrets/immutable.yaml
```

internal/secrets/config.go
ralgozino: Fixed. It now resolves through the same parserx.ConfigParser the configuration loader uses, so provider.type: "{env://AUTH_TYPE}" with AUTH_TYPE=sso selects the four Pomerium secrets — verified.

Where a value can't be resolved, the command now stops and names the field rather than reading it as empty. That matters most for the etcd encryption field: an unfollowable {file://} there looks like a cluster with no encryption at rest, and furyctl would write a second key.

internal/secrets/secrets.go
stefanoghinelli: Empty files and dir look ok. There might still be a case of a valid but unreadable file (for example, due to incorrect permissions); I don't know if you've had a chance to test this.

stefanoghinelli (commented):
Just 1 point, then LGTM

internal/secrets/secrets.go
ralgozino: A don't think that is a real concern, you will realistically use `create secrets` either when you are creating the cluster configuration or when enabling a feature like SSO on an existing cluster.

In both cases, it is safe to assume that you can access all the needed files IMO.

Having said that, better safe than sorry so I updated the logic in https://github.com/sighupio/furyctl/pull/756/commits/43051d1f1d6fc81281658e515d1b7a7cf4eb29a8 for that case too:

Now a file that cannot be accessed is an error, like the empty file and the directory — all three are the same question, and only the error variant avoids handing you a configuration that fails later:

```
ERRO the path holds no secret: furyctl cannot read the file .../basic-auth/password:
     open .../basic-auth/password: permission denied
```

Exit 1, no reference written, and the file's value untouched. With the permissions restored the run reports it as kept and writes the reference as before. Covered by a test row next to the other two, which skips itself when the account can read a 000 file, since CI runs as root.

stefanoghinelli (approved):
🚀

PR #760, opened 2026-09-14
chore(immutable): drop the node arch backfill

ralgozino:
### Summary 💡

Removes the hand-written `arch` default for Immutable nodes. The distribution now requires the field, so schema validation rejects a configuration without it before this code runs.

Relates:
- sighupio/distribution#573
- sighupio/distribution#596

### Description 📝

`rawNodesByHostname` wrote `x86-64` into any node that omitted `arch`, because the butane templates read that value without a guard. Its comment states the underlying problem plainly: furyctl does not apply JSON-schema defaults, so the `"default": "x86-64"` the distribution declared was never applied anywhere.

The distribution now makes `arch` required and `furyctl create config` writes it on every node, so the backfill is unreachable.

It was never sufficient anyway. It covered only the butane path; a node without `arch` still stopped the apply in the kubernetes phase (`hosts.yaml.tpl`) and in the infrastructure phase's ansible inventory, and `extractUsedArchitectures` still produced an empty architecture. No cluster ever reached the point where the backfill helped.

`TestRawNodesByHostnameArchDefault` guarded the removed behaviour and goes with it.

### Breaking Changes 💔

None.

### Tests performed 🧪

- [x] Built furyctl from this branch and ran `furyctl create config --kind Immutable` followed by `furyctl apply --dry-run` against the matching distribution branch: the generated configuration renders the infrastructure and kubernetes phases with no backfill in play.
- [x] Verified that an apply of a configuration file that does not specify a node's architecture now fails at validation time saying the field is missing.
- [x] Verified a correct configuration file still runs a full apply without errors.

### Future work 🔧

None.

- ~~[ ] I've updated the `docs/releases/unreleased.md` file (or equivalent)~~ not applicable, no user-facing change
- [x] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change — the code is under `internal/apis/kfd/v1alpha2/immutable/` and reaches no other kind
- [x] CI is green


stefanoghinelli (approved):
lgtm

PR #762, opened 2026-09-17
fix(cache): do not serve a moving git ref from the local cache

ralgozino:
### Summary 💡

furyctl cache did not invalidate when downloading from a branch and that branch has received commits since the last download. Now furyctl will invalidate the cache if the branch has received new commits and do the download.

### Description 📝

The local download cache keys its entries on the URL, and a git URL carries its ref as a query parameter. For a tag or a commit SHA the URL identifies the content, so a cache hit is always correct. For a branch it does not: two commits of one branch give the same URL, so the first download of that branch is served for ever and a later push is invisible.

Anyone who pins an installer or the distribution to a branch therefore keeps the copy from the first run, whatever they push afterwards. The vendor folder is rewritten from the stale cache on each run, so its timestamp changes and the staleness is hard to see.

Serve the cache only when the ref always resolves to the same commit: a version tag or a full commit SHA. For any other ref, and for a URL with no ref, remove the cache entry and the destination, then download again.

### Breaking Changes 💔

None - previous behaviour was a bug IMO.

### Tests performed 🧪

- [x] Tested that cache is invalidate when a new commit lands on a branch (for example pointing to a branch for a module/installer on kfd.yaml)

### Future work 🔧

None

> [!IMPORTANT]
>

- [x] CI is green


stefanoghinelli (approved):
Looks good

PR #764, opened 2026-09-18
feat(immutable): loadbalancers upgrade

ralgozino:
### Summary 💡

The Immutable infrastructure phase now upgrades the load balancers. Before this change it returned at once and logged "not implemented yet", so an upgrade left the load balancers with the operating system and the extensions of their installation.

Relates: sighupio/distribution#601, sighupio/installer-immutable#26

### Description 📝

Two commits. The first one corrects a panic that is present today, and the second one adds the feature. Each one builds and passes the tests on its own.

**`fix(immutable)`: give the run every phase of a stored upgrade state.** Every field of `upgrade.Phases` is an optional pointer, so a stored state holds only the phases that the run tracked, and a write to a phase that the state does not hold stops furyctl with a panic. `apply --upgrade --phase infrastructure` stores a state of that one phase, and a later full apply then writes `PreKubernetes`, which that state does not hold.

`readUpgradeState` reads the stored state twice, because the two answers need two different readings of it. The phase to resume from needs the phases exactly as the earlier run stored them: `GetLatestResumablePhase` skips a phase that the state does not hold, and a phase that this version adds must stay absent for that decision. The run itself reads the state over `initUpgradeState`, so every phase exists.

**`feat(immutable)`: upgrade the load balancers.** `Infrastructure.Exec` returned before it rendered anything during an upgrade, so the phase folder held no inventory and no playbook. The rendering moves to `renderAnsible`, which both paths call. Only the create path runs `BootstrapNodes` and the iPXE server, because an upgrade never re-provisions a host. `upgradeLoadBalancers` then runs the `pre-infrastructure` hook, the playbook and the `post-infrastructure` hook, and records the three infrastructure phases in the upgrade state.

**Older distributions.** The infrastructure phase now runs on every upgrade, and a distribution released before this feature gives no `upgrade-load-balancers.yml`. An upgrade from 1.35.0 to 1.35.1 stopped with "the playbook could not be found". `loadBalancerPlaybookShipped` reads the templates of the distribution, and the phase writes a warning and continues when the playbook is absent.

The test reads the distribution and not the folder of the phase. A render writes the files of the distribution over that folder and removes nothing, so a folder that an earlier run left with a playbook makes an older distribution read as one that gives it. A lab run found this: the folder held `hosts.yaml` of 1.35.1 next to the playbook of a 1.36.0 run, and ansible stopped with `missing required arguments: os_update_target_version`.

**`--upgrade-node` routes on the role.** `upgradeNodeRole` reads the role with `NodeRole()`, which the schema already gives:

| Role of the host | Result |
|---|---|
| `worker` | The kubernetes phase runs `upgrade-worker-nodes.yml --limit <host>`, as before. |
| `loadbalancer` | The infrastructure phase runs the load balancer playbook, then the other phases do not run. |
| `controlplane`, `etcd` | Rejected. These hosts are upgraded as a group, with the quorum guarded. |
| not in the configuration | Rejected. |

The test runs before any phase, so a wrong name stops the command at once. Before this change furyctl accepted any string, and a load balancer name reached a playbook that targets the `nodes` group, so nothing was upgraded.

**The target is a variable, not `--limit`.** `keepalived.conf` renders a unicast peer list from every load balancer, so the playbook gathers the facts of the whole group in a first play. `--limit` filters that play too, so the target goes in as `-e lb_upgrade_target=<host>`.

**Other kinds.** Every change is inside the Immutable provider. The rule that `--upgrade-node` selects its own phase belongs to this kind alone, because only here does the phase depend on the role. OnPremises upgrades every `--upgrade-node` host in its kubernetes phase and keeps its behavior. The help text of `--skip-nodes-upgrade` and `--upgrade-node` changes for every kind: both said "On kind OnPremises", and the first one also said that only the control plane is upgraded, which was never true. etcd is always upgraded on OnPremises and on Immutable.

### Breaking Changes 💔

Not an actual breaking change because previous behaviour was a bug IMO. But just to be clear:

`apply --upgrade-node <host>` on kind Immutable now stops when you give `--phase`, `--start-from` or `--post-apply-phases` with it. Before this release furyctl accepted them together. `--phase kubernetes` did the same work as `--upgrade-node` alone, and the other two did no extra work and said nothing. Use `--upgrade-node` alone. Kind OnPremises is not affected.

### Tests performed 🧪

- [x] Upgraded each load balancer of a two-node lab in turn with `apply --upgrade-node <host>`: Flatcar 4593.2.3 to 4757.2.0, containerd 2.3.2 to 2.3.5, keepalived v2.3.4 to v2.4.3, flatcar-python 4593.2.3 to 4757.2.0
- [x] Verified that the host that is not the target only gathers facts, which shows that the variable selects one host while the group stays in the play
- [x] Upgraded both load balancers in one run with `apply --upgrade --phase infrastructure`, which also runs the two hooks and writes the upgrade state
- [x] Verified that the virtual IP address moved to the peer during each reboot, with one holder and no split brain after the run
- [x] Powered off one load balancer and confirmed that the run stops before it stages anything, and that the host that answered kept its versions, its uptime and its virtual IP address
- [x] Upgraded a 1.35.0 cluster to 1.35.1 with released distributions and confirmed the warning, no ansible play, and a successful run
- [x] Confirmed that same run skips although the folder of the phase still held a playbook of an earlier 1.36.0 run
- [x] `TestUpgradeNodeRole`, `TestValidateUpgradeNodeRejectsAPhaseSelection`, `TestReadUpgradeState`, `TestInitUpgradeStateCoversEveryPhase` and `TestLoadBalancerPlaybookShipped` cover the decisions above

### Future work 🔧

- `Infrastructure.Exec` ignores its `startFrom` parameter, so the phase has no `--start-from` support. It also keeps no per-host state, so an interrupted rollout of every load balancer starts again. Both match the other phases today.
- `--skip-nodes-upgrade` keeps its meaning of "worker nodes only" and does not skip the load balancers. A flag for them waits for the staged worker bookkeeping.
- `CopyFromTemplate` removes nothing from the folder of a phase, so a file of an earlier run stays there. This change reads the distribution to avoid it. The correction of the shared operation needs a pull request that can test every kind.
- The fields of `upgrade.Phases` are optional pointers, and an absent phase carries a meaning that two functions read. A value type with an empty status as the sentinel removes the panic for every kind. That also needs a pull request that can test every kind.
- The `keepalived` role does not run in `upgrade-control-plane.yml`, so a cluster that runs keepalived on the control planes keeps its installed version.

- [x] My PR has a clear scope and does not mix together several unrelated changes: Almost, this could have been splitted into two (the first fix on the panic and the rest) but they were to close together.
- [X] I've considered all the different cluster kinds (KFDDistribution, OnPremises, EKSCluster, Immutable) that may be affected by this change
- [x] CI is green


stefanoghinelli (approved):
lgtm

